PR #154, opened 2021-01-19
New web-money block for K-54

lukestanley:
Adds a new 'web-money' block, and makes player pause and play "sub-block" events work properly.

It can be used like this:
```javascript:
{
  "title": "Basic Player Web Money",
  "blocks": [
    {
      "type": "init"
    },
    {
      "type": "mapping",
      "mapping": "\n    {\n  \"display_artist\": 'Artist name',\n  \"display_title\": 'Track title',\n  \"url\": 'track.mp3',\n  \"paymentPointer\":'$ilp.example.com/some_payment_pointer}\n"
    },
    {
      "type": "player",
      "titleMapping": "data && data.join(' - ', [display_artist, display_title])",
      "onPlay": [
        {
          "type": "web-money"
        }
      ],
      "onPause": [
        {
          "type": "web-money",
          "enabled": false
        }
      ]
    }
  ],
  "id": "basicPlayerWebMoney",
  "adapterName": "examples"
}
```
Part of https://linear.app/kendraio/issue/K-54/demo-playing-a-track-and-tipping

### DIFF
--- a/src/app/blocks/audio-player-block/audio-player-block.component.html
+++ b/src/app/blocks/audio-player-block/audio-player-block.component.html
@@ -20,12 +20,12 @@
                        [models]="[model]"
                        [context]="context"></app-blocks-workflow>
 </div>
-<div *ngIf="!hasError && onPlayBlocks.length > 0">
+<div *ngIf="!hasError && onPlayBlocks.length > 0 && isPlaying">
   <app-blocks-workflow [blocks]="onPlayBlocks"
                        [models]="[model]"
                        [context]="context"></app-blocks-workflow>
 </div>
-<div *ngIf="!hasError && onPauseBlocks.length > 0">
+<div *ngIf="isPlaying == false && onPauseBlocks.length > 0">
   <app-blocks-workflow [blocks]="onPauseBlocks"
                        [models]="[model]"
                        [context]="context"></app-blocks-workflow>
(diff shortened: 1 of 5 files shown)


src/app/blocks/audio-player-block/audio-player-block.component.html
```diff
@@ -20,12 +20,12 @@
...
                        [context]="context"></app-blocks-workflow>
 </div>
-<div *ngIf="!hasError && onPlayBlocks.length > 0">
+<div *ngIf="!hasError && onPlayBlocks.length > 0 && isPlaying">
   <app-blocks-workflow [blocks]="onPlayBlocks"
                        [models]="[model]"
                        [context]="context"></app-blocks-workflow>
 </div>
-<div *ngIf="!hasError && onPauseBlocks.length > 0">
+<div *ngIf="isPlaying == false && onPauseBlocks.length > 0">
```
lukestanley, line 28: I think pause events need to happen even if there is an error, if a user cannot play something e.g: due to network error, they might not be happy to keep paying, so those events need to keep flowing.

lukestanley: I think the onPauseBlock's should only exist if isPlaying returns false, hope you agree this is a sensible pattern for doing it.

darrenmothersele: agreed! nice one

darrenmothersele (approved):
nice one!

darrenmothersele (approved):

Issue #159, opened 2021-02-05
Coil login button (K-70)


lukestanley:
We need a Coil login button as part of integrating the Web Monetization prototype, for users without the Coil extension (including mobile users).

https://docs.google.com/document/d/1e76N05aLZ9URaCndBB53ryDwbo9o3p4qZfpVNqrc-UY/edit#heading=h.ft64zgwshbm
https://linear.app/kendraio/issue/K-70/coil-login-button
https://linear.app/kendraio/issue/K-53/integrating-web-monetization-prototype

lukestanley:
WIP: 
https://github.com/kendraio/coil-api-proxy 
https://github.com/kendraio/kendraio-app/tree/k-70-coil-login-button

lukestanley:
Done

PR #160, opened 2021-02-12
K-70 Coil login button

lukestanley:
Implements the frontend part of https://github.com/kendraio/kendraio-app/issues/159

### DIFF
--- a/src/app/blocks/web-money/web-money.component.ts
+++ b/src/app/blocks/web-money/web-money.component.ts
@@ -33,11 +33,67 @@ function removePaymentPointer() {
   }
 };
 
+const COIL_CLIENT_ID = '19d9c8e7-3a37-4fd5-803c-057300f4354b';
+function uuid4() {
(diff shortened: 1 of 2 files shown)


src/app/blocks/web-money/web-money.component.ts
```diff
@@ -33,22 +33,115 @@ function removePaymentPointer() {
   }
 };
 
+const COIL_CLIENT_ID = '19d9c8e7-3a37-4fd5-803c-057300f4354b';
+function uuid4() {
```
darrenmothersele, line 37: we have the UUID library already as a dependency, you could use that to generate this uuid if you want instead of including uuid generating code. Something like:
```js
import uuid from 'uuid';
console.log(uuid.v4());
```

darrenmothersele (approved):
Thanks!

src/app/blocks/web-money/web-money.component.ts
lukestanley: Done, thanks. Will merge now.

Issue #163, opened 2021-02-25
Refresh redirect URL after Web Money events (K-91)


lukestanley:

lukestanley:
Fixed with above merge.

PR #164, opened 2021-02-25
Generates login URL (with redirect) after config event, copy fixes and readme improvement 

lukestanley:
- This fix ensures the latest app URL is used for the redirect back from Coil. (K-91, #163)

- The player description no longer describes the Web Monetization support as "in development". (K-90)

-  "No Web Monetization extension found 😟." is no longer shown since we have a solution with Coil login. (K-89)

-  Adds developer setup documentation to README (from Google Doc)


darrenmothersele (approved):
nice one

Issue #175, opened 2021-03-11
Capture track play statistics


lukestanley:
We'd like to provide artists play and payment statistics per track by locally aggregating and sending non-personal session information to an endpoint (and database) defined using block configuration.

We should allow periodically sending information such as:

- a timestamp
- current Audio track time position
- Web Monetization track payment total (if any)
- Web Monetization track payment asset code (e.g: 'xpr')
- Web Monetization payment pointer (for easy lookup and troubleshooting)
- Web Monetization session payment total (per asset code)
- The user status: (anon or logged in - but no user identifier) 
- track ID,
- a non-persistent session UUID (to track the current session, with no tracking linkable across sessions).

While playing or streaming payments, to conserve resources, analytics should be sent in batches e.g: no more than one web request every 10 seconds while playing, but still allowing uploading a majority of data even if resources are suddenly unavailable - such as suspending, low memory event etc).

lukestanley:
The above has been implemented, tracking this subset:

-  a timestamp
-  Web Monetization track payment total (if any)
-  Web Monetization track payment asset code (e.g: 'xpr')
-  Web Monetization session payment total (per asset code)
-  track ID
-  a non-persistent session UUID (to track the current session, with no tracking linkable across sessions).


It does **_not yet track_**:

- current Audio track time position
- Web Monetization payment pointer (for easy lookup and troubleshooting)
- The user status: (anon or logged in - but no user identifier)

PR #176, opened 2021-03-25
K-108 / #175 - Tracking play statistics

lukestanley:
A fairly generic tracking module is added that sends a Graphql instance data no more than once every 10 seconds.
It is used by Web Money when tracking is enabled via block config.

It may be tested by viewing the "player/all" flow.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10898,6 +10898,11 @@
       "resolved": "https://registry.npmjs.org/json-stringify-safe/-/json-stringify-safe-5.0.1.tgz",
       "integrity": "sha1-Epai1Y/UXxmg9s4B1lcB4sc1tus="
     },
+    "json-to-graphql-query": {
+      "version": "2.0.0",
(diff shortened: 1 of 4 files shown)


darrenmothersele (approved):
what happened to the semicolons in tracking.ts?

lukestanley:
It never had any semicolons in tracking.js :)
It's a new class so I had no existing style to maintain.
If you feel it's less readable, or if we have a style guide that is pro-semicolons then I'm happy to adhere to it :)

I should probably remove the console.log's and use console.info selectively though.

darrenmothersele:
We don't have a JS style guide, but I thought we had linting rules in place? Maybe we should be explicit about it. Especially when we have more developers potentially joining. 

I know semicolons are optional in JS, but there are rare occurances that ASI (automatic semicolon insertion) can cause weird bugs. I don't see any reason not to use them.

darrenmothersele (approved):

PR #185, opened 2021-04-30
Icons and text of blocks

lukestanley:
Adds icons to the list of blocks in the workflow sidebar.
When adding a block task, you can now see text in the block modal without having to mouseover.


darrenmothersele (approved):
nice one!

PR #197, opened 2021-07-07
Adds Firebase headers for CSP and XSS restrictions

lukestanley:


lukestanley:
I would prefer testing this without merging into main but the setup to do that correctly would likely take much more effort than reverting and redeploying if this is broken.
How does this look to you @darrenmothersele?

darrenmothersele (approved):

PR #203, opened 2021-07-16
flows api function 

the-greenman:
This code adds a new API endpoint "/flows". 
This endpoint creates a complete dump of all flows. Output is essentially a combination of the base "/api" endpoint, and the individual "workflow" endpoint. 

The initial use of this endpoint is to enable a deep search of all flows and their content. 

It also includes @lukestanley's updates to the headers for CSP and XSS

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -23,7 +23,15 @@
         "source": "**",
         "function": "dynamicIndex"
       }
-    ]
+    ],
+    "headers": [{
+      "source": "**/*",
+      "headers": [
+        {"key": "Content-Security-Policy", "value": "default-src 'self'; script-src cdn.coil.com www.google-analytics.com ajax.googleapis.com maps.googleapis.com youtube.com s.ytimg.com 'self' 'unsafe-eval'; style-src https://fonts.googleapis.com https://fonts.gstatic.com 'self' 'unsafe-inline'; img-src *; frame-ancestors 'none'; connect-src *; frame-src ajax.googleapis.com maps.googleapis.com youtube.com s.ytimg.com; font-src 'unsafe-inline' data: https://fonts.googleapis.com https://fonts.gstatic.com 'self'; form-action 'self';"},
+        {"key": "X-Frame-Options", "value": "DENY"},
+        {"key": "X-XSS-Protection", "value": "1; mode=block"}
+      ]
+    }]
   },
   "functions": {
     "predeploy": [
--- a/firebase.json
+++ b/firebase.json
@@ -16,6 +16,10 @@
         "function": "api"
       },
       {
+        "source": "/flows",
+        "function": "api"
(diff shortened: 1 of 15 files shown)


functions/src/handlers/workflow-cloud.ts
```diff
@@ -153,6 +153,27 @@ function appFactory({db, auth}: { db: any, auth: admin.auth.Auth }) {
     res.send(result);
   });
 
+  app.get('/flows/', async (req, res) => {
+    const docs = await db.collectionGroup('workflows').get();
+    const result: Array<any> = [];
+    docs.forEach(doc => {
+      let data = doc.data();
+      let unEncode = JSON.parse(data._encoded);
```
lukestanley, line 161: Why use ._encoded and have to decode it ourselves?
I suspect  `data()` should be used instead of having to decode it.
I know it matches a pattern from the existing code but I suspect the existing code should not be using internal objects either (internal as marked by the `_` prefix.
I think this is a relevent example:
https://firebase.google.com/docs/firestore/query-data/get-data#get_multiple_documents_from_a_collection
I assume it can return a regular dictionary object in the same way.
Behaviour of internal objects may change, I couldn't find _encoded in a quick search of the Firebase documentation.

lukestanley (changes requested):

functions/src/handlers/workflow-cloud.ts
the-greenman: The _encoded is actually the name of the field. We're storing the flow as a single JSON "encoded" string in the Firebase document.

darrenmothersele:
_encoded is the field name. There used to be a bug when saving JSON objects with arrays directly to firestore. Arrays of nested objects would come back in the wrong order. This meant that flows would get corrupted because blocks could get reordered. Saving the full serialised text was a work around. The content of the flow didn't need indexing so wasn't an issue. Metadata for indexing is added as separate fields alongside the _encoded field.

functions/src/handlers/workflow-cloud.ts
darrenmothersele: Comment below hopefully explains history of this

lukestanley: Ahh okay! That changes my view of this then.

firebase.json
```diff
@@ -15,16 +15,42 @@
...
+      {
+        "source": "/flows",
+        "function": "api"
+      },
       {
         "source": "**",
         "function": "dynamicIndex"
       }
-    ]
+    ],
```
lukestanley, line 26: I wasn't expecting to this added as a new thing, I assume GitHub or Git is acting weird.

lukestanley (approved):
Now I know it's a field name 👍

PR #212, opened 2021-10-01
The-greenman/issue207

the-greenman:
New updates to grid to enable passthrough and automatic column definitions.

### DIFF
--- a/src/app/blocks/grid-block/grid-block.component.ts
+++ b/src/app/blocks/grid-block/grid-block.component.ts
@@ -1,9 +1,9 @@
-import {Component, ElementRef, EventEmitter, Input, NgZone, OnChanges, OnInit, Output, ViewChild} from '@angular/core';
-import {clone, get, has, isArray, isObject} from 'lodash-es';
-import {search} from 'jmespath';
-import {WorkflowCellRendererComponent} from '../../components/workflow-cell-renderer/workflow-cell-renderer.component';
-import {mappingUtility} from '../mapping-block/mapping-util';
-import {ConnectionStatusRendererComponent} from '../../components/connection-status-renderer/connection-status-renderer.component';
+import { Component, ElementRef, EventEmitter, Input, NgZone, OnChanges, OnInit, Output, ViewChild } from '@angular/core';
+import { clone, get, has, isArray, isObject } from 'lodash-es';
+import { search } from 'jmespath';
+import { WorkflowCellRendererComponent } from '../../components/workflow-cell-renderer/workflow-cell-renderer.component';
+import { mappingUtility } from '../mapping-block/mapping-util';
+import { ConnectionStatusRendererComponent } from '../../components/connection-status-renderer/connection-status-renderer.component';
 
 @Component({
   selector: 'app-grid-block',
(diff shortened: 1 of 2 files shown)


src/app/blocks/grid-block/grid-block.component.ts
```diff
@@ -55,9 +57,16 @@ export class GridBlockComponent implements OnInit, OnChanges {
   }
 
   updateOutputDisplay() {
-    const defaultCols = isArray(this.model) && this.model.length > 0
-      ? Object.keys(this.model[0]).map(key => ({ headerName: key, field: key }))
-      : [];
+    let defaultCols = [];
```
the-greenman, line 60: I've expanded this to a more verbose form for readability.

src/app/blocks/grid-block/grid-block.component.ts
```diff
@@ -55,9 +57,16 @@ export class GridBlockComponent implements OnInit, OnChanges {
...
+    if (isArray(this.model) && this.model.length > 0) {      
+      defaultCols = Object.keys(this.model[0]).map((key) => {
+          let column = { headerName: key, field: key };
+          if (this.firstRowHeaders) column.headerName = this.model[0][key];
+          return column;
+        }        
+      )
+      if (this.firstRowHeaders) this.model.shift(); // remove first row
```
lukestanley, line 68: Is it more safe for this.model be explicitly set to a new value?
I wonder if mutating the model directly is the right thing to do, vs explicitly setting the model because we already have issues with how 'reactive' it is? :thinking:

src/app/blocks/grid-block/grid-block.component.ts
```diff
@@ -78,32 +87,36 @@ export class GridBlockComponent implements OnInit, OnChanges {
...
+        ? { cellRendererParams: { ...item.cellRendererParams, context: this.context } }
         : {}
     }));
   }
 
   onSelectionChanged(e) {
     // console.log({ e });
     const selectedRows = e.api.getSelectedRows();
-     // console.log({ selectedRows });
+    // console.log({ selectedRows });
```
lukestanley, line 119: maybe we should delete these.

src/app/blocks/grid-block/grid-block.component.ts
the-greenman: I'll switch to a clone, and not alter the underlying model here. This current implementation will cause the removal of the first row to be passed on through the chain.

src/app/blocks/grid-block/grid-block.component.ts
the-greenman: All old comments deleted.

src/app/blocks/grid-block/grid-block.component.ts
```diff
@@ -64,32 +57,28 @@ export class GridBlockComponent implements OnInit, OnChanges {
           if (this.firstRowHeaders) column.headerName = this.model[0][key];
           return column;
         }        
-      )
-      if (this.firstRowHeaders) this.model.shift(); // remove first row
+      )      
     }
+    let workingModel = this.model; // create a local clone of the data - to allow us to modify data only for display
+    if (this.firstRowHeaders && workingModel.length >0) workingModel.shift(); // remove first row from the working model
```
the-greenman, line 63: I am now cloning the model to a local var. Original data is untouched.

lukestanley: That makes sense but there is an actual clone method BTW that lodash has, which is already imported (and used below).

the-greenman: Now using "clone". Not just pretending.

lukestanley (approved):

PR #215, opened 2021-10-12
Installs Localize, enables Ivy & Localize

lukestanley:
This now can compile and the app seems to once again load properly, however there are logs warning about issues, and it has not been tested on many flows.


the-greenman:
I think the only immediate thing to do is probably to actually bump up the minor version number here.

the-greenman (approved):
Looks good

PR #226, opened 2021-11-03
Angular 12 upgrade

lukestanley:
Automated and manual migrations and version bumps have been applied.

Compatible security updates have been too, and an update to @angular-devkit/build-angular is expected to fix more soon.

**Please change the target branch to develop after resolving the existing Angular PR's.**

(It is currently pointing to the prior version as the target, to show the difference between the last version and Angular 12 branch.)


the-greenman (approved):
Routing seems to still be working, and stylesheets seem Ok. 
No specific reason to not let this through.

PR #227, opened 2021-11-04
Angular 13

lukestanley:
Angular 13 just got released!
Still awaiting an upgrade of ansi-regex. I tried using npm-force-resolutions but there is a breaking change for how it is imported, at a level we don't control.
flex-layout have not released a new verison yet, so I have skipped upgrading this dependency and it still works fine.

- [x] **Don't forget to merge to develop, not angular-12!**


the-greenman:
Testing CSV import: 
Excel files are importing fine. 

We do get an error popping up when the page loads: Cannot read properties of undefined (reading 'Symbol')
Not immediately able to find the source...

the-greenman:
I merged the latest code from dev to help with testing functionality.

the-greenman:
https://kendraio-app-e3dsc3bw4-kendraio.vercel.app/kendraio/coronavirus

TypeError: sort_by() expected argument 1 to be type array but received type null instead.

Leaflet maps are not showing points.

the-greenman:
Far less build warnings in this version... but we can probably still remove all the code that was being warned about previously.

the-greenman:
The covid map issue is data related. The API has changed. Map is working.

lukestanley:
Yeah, looking at https://app.kendra.io/kendraio/coronavirus and https://kendraio-app-e3dsc3bw4-kendraio.vercel.app/kendraio/coronavirus
Maps, charts, and everything else seem the same.
Have you found any actual regression? @the-greenman

the-greenman:
> Yeah, looking at https://app.kendra.io/kendraio/coronavirus and https://kendraio-app-e3dsc3bw4-kendraio.vercel.app/kendraio/coronavirus Maps, charts, and everything else seem the same. Have you found any actual regression? @the-greenman

So far, no actual regression. As you noted, everything I've tested has worked. All of what I thought were new errors have been appearing in the current production app.

lukestanley:
I've noticed @angular-devkit/schematics is a new non dev dependency. I think this is a mistake, I will try moving it.

the-greenman (approved):
There is no code here that is an issue. We've not found any problems in the flows. 
I think getting dev to v13 is a very important step and we should take it.

PR #229, opened 2021-11-09
Whitelist Angular onload style change script

lukestanley:
A stylesheet aimed to set it's media type to all useragent types, rather than being print specific by doing this inline script after it loaded:
"this.media='all'"

It failed due to the Content-Security-Policy, but by hashing the inline script, and adding it to the allow list, I hope to remedy the issue and fix the CSS.

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -27,7 +27,7 @@
     "headers": [{
       "source": "**/*",
       "headers": [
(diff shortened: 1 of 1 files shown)


the-greenman (approved):
Lets try it. I'd like to know what the cause of this is. But lets see if this works first.

lukestanley:
It happend silently during the Angular 12 upgrdade, more info on it here:
https://stackoverflow.com/questions/67565858/angular-12-css-optimization-inline-event-handler-with-content-security-policy

PR #230, opened 2021-11-10
Angular CSP fix: disabling inline optimizations

lukestanley:
Though the CSP issue seems resolved in Chrome and Firefox, Safari/WebKit says:`Refused to execute a script for an inline event handler because 'unsafe-inline' does not appear in the script-src ` and refuses to show the page correctly unless inline optimizations are turned off.
So this turns them off.


the-greenman:
Let's get back up and running, and then you can explain to me what the real causes are.

the-greenman (approved):

PR #235, opened 2021-11-12
Read the docs config

the-greenman:
I've now created a default RTD config file, and pinned to an older version of the docutils.  

Doc update docs to remove all sphinx warnings.


lukestanley:
I see the pinning and I get that, what are the documentation changes about? @the-greenman

lukestanley:
e.g why "Removed complex form example"? @the-greenman

docs/workflow/blocks/form.rst
```diff
@@ -80,64 +80,4 @@ A simple search form without a submit button.
         }
     },
     "uiSchema": {}
-  }
```
lukestanley, line 83: I didn't get the purpose of removing this yet and just wanted to check.

lukestanley (approved):

the-greenman:
Ah. I actually thought those had already been merged! 
I went through all the docs and removed all the warnings from Sphinx.

docs/workflow/blocks/form.rst
the-greenman: I'll bring the complex example back. I was just removing all the updates added at the time of the break.

PR #238, opened 2021-11-13
CI with E2E and Unit tests - Swaps Protractor for Cypress, uses Karma Electron

lukestanley:
**A GitHub Action now automatically runs all E2E tests and unit tests.**

This PR replaces Angular Protractor (which is depreciated) with Cypress (the most popular alternative apparently).
It adds some new E2E tests and fixes an old broken one.

Please see the readme doc outlining the 3 new E2E testing commands.
https://github.com/kendraio/kendraio-app/pull/238/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5

`npx ng test` runs unit tests with Electron right away.
I added a new unit test for UUID hashing as used by the mapping block, mirroring the E2E test.

In addition to replacing Protractor with Cypress, some dev build speed improvements were made by disabling minify for the dev build, turning off "extractLicenses", "buildOptimizer", and "optimization" for fonts and scripts.
The defaultConfiguration was set to dev.

**Note on testing CI integration experiment**

To validate how new end-to-end testing might work in CI, I temporarily modified the Vercel build command to:
```
yum install -y xorg-x11-server-Xvfb gtk2-devel gtk3-devel libnotify-devel GConf2 nss libXScrnSaver alsa-lib;
npm run e2e:serve_and_test && ng build
```

and tried it on my new E2E branch.
It successfully results in preventing deployment when tests are failing, but I estimate that it adds 2-3 minutes to the Vercel build time. Using another service with the site generated from the build could be faster and less of an abuse extension of the build pipeline.

See the deliberate regression of the UUID function being detected: https://vercel.com/kendraio/kendraio-app/4o9AsiyofqvUbQq6pWKnpuXpNgME
`15:29:38.299 "AssertionError: Timed out retrying after 10000ms: Expected to find content: '3ab8d0cd-7b76-5741-8bc9-5725650dc435' but never did."`
On another build without the deliberate regression:
https://vercel.com/kendraio/kendraio-app/7kuwY6KvqSd7GB4RGvoFmHcenUTT
`15:04:45.115 "✔ All specs passed!"`
Everything passed in the browser tests and it then continues to deploy, though maybe 3 minutes longer than normal.


the-greenman (commented):
No real comments at a code level on this. I think the best way forward will be to write some real tests and try the whole process. I'm working on some tests for my new State service.

the-greenman (approved):
This is good to move to dev.

PR #241, opened 2021-11-28
State: Global and Local

the-greenman:
This branch adds a new Shared State service, and updates a number of blocks to interact with it. 

The Context (Save) Block, now called "Context & State (Save)" block can now write to the state. 
The Mapping block, and others, has been updated to read from the state block. 

The state is a shared data store that provides some blocks with asynchronous access to data. This capability should never modify the data emitted by the block, as this could easily result in infinite loops within out flows. It can however be used in elements of the UI to present up to date information to the user, or to enable/disable components.


src/app/blocks/context-save-block/context-save-block.component.ts
```diff
@@ -11,28 +12,52 @@ import {mappingUtility} from '../mapping-block/mapping-util';
 })
 export class ContextSaveBlockComponent extends BaseBlockComponent {
 
-  contextKey = 'saved';
+  key = 'saved';
```
the-greenman, line 15: I've changed the key from "contextKey" to "key" because we are no longer  just dealing with context. 
The old contextKey will still work for existing blocks.

src/app/blocks/debug-block/debug-block.component.ts
```diff
@@ -17,21 +17,28 @@ export class DebugBlockComponent implements OnInit, OnChanges {
   @Output() output = new EventEmitter();
 
   open = 1;
+  showData = true;
```
the-greenman, line 20: New config to standardise what to show: data, context or state

src/app/services/shared-state.service.ts
```diff
@@ -0,0 +1,93 @@
...
+ * To add a shared state capabilities to a component, you need to subscribe to updates in the constructor. 
+ * 
+ * @example
+ *   constructor(
+ *    private stateService:SharedStateService
+ *    ) {
+ *        stateService.shared$.subscribe(state => { setTimeout(() =>{this.updateOutputDisplay()}) });
+ *       }
+ * 
+ *  The setTimeout() wrapper allows the updates to work with Angular's change detection cycle. Don't leave it out.
```
lukestanley, line 16: I'm curious, without setTimeout, what happens? Are ExpressionChangedAfterItHasBeenCheckedError errors raised?
I'm not an Angular expect, but I wonder if a better solution would be to do the subscription at a different stage of the component's lifecycle (if possible?) e.g: ngAfterContentChecked or ngAfterContentInit vs in the constructor?
@the-greenman

the-greenman: Yes, the setTimeout deals with the ExpressionChangedAfterItHasBeenCheckedError. 

I'll see if there are other possible subscription locations.

the-greenman: Doesn't look like the location of the subscribe matters. It just sets up the listener, which will then get called when data arrives.

the-greenman: Hmm. Maybe I'm lying. ngAfterViewInit is working in my initial test

the-greenman: nope.

the-greenman: I've attempted to use the onPush change detection strategy. And then call "this.cd.markForCheck()" when I have an update. But still get the issue. 

I've tried the subscribe in multiple locations.

the-greenman: https://hackernoon.com/everything-you-need-to-know-about-the-expressionchangedafterithasbeencheckederror-error-e3fd9ce7dbb4

Here is a handy writeup. I think the underlying challenge is that our flows are using the main angular change detection. So, the "correct" way to do this is to pass the changes down through the flow again.  But this is not what we want. 

The solution that looks viable from this article (one of many I've read) seems to be to create a microtask with a promise, rather than a macrotask with the timeout.

cypress/integration/context_and_state.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  it('should be allow a template block to be able to access global state values', () => {
+    loadFlowCode([     
+      {
+         "type":"template",
+         "template":"{{state.global.test}}|{{state.local.test}}"
+      },
+      {
+        "type": "context-save",
+        "valueGetter": "uuid('global')",
```
lukestanley, line 22: Why not just set a string directly? I don't think the UUID method helps.

src/app/blocks/context-save-block/context-save-block.component.ts
```diff
@@ -11,28 +12,52 @@ import {mappingUtility} from '../mapping-block/mapping-util';
...
+
+  constructor(private stateService: SharedStateService) {
+    super()
+  }
+
 
   onConfigUpdate(config: any) {
-    this.contextKey = get(config, 'contextKey', 'saved');
+    if (has(config,'key')) {
+      this.key = get(config, 'key', 'saved');
```
lukestanley, line 28: if they have both maybe we should warn them?

cypress/integration/context_and_state.ts
lukestanley: I think the way you do it below is more simple.

cypress/integration/context_and_state.ts
```diff
@@ -0,0 +1,97 @@
...
+    },
+    {
+      "type": "grid",
+      "gridOptions": {},
+      "passThrough": false,
+      "firstRowHeaders": true,
+      "enabledGetter": "state.global.disabled"
+    }
+  ]);  
+  cy.get(".ag-root",{ timeout: 10000 }).should('not.exist');
```
lukestanley, line 93: I only see this one assertion. Above it says, 'should allow the enabling and disabling of the datagrid'

lukestanley, line 93: What is the .ag-root class?
perhaps a variable like DATAGRID_SELECTOR could clarify?

docs/workflow/blocks/context_block.rst
```diff
@@ -1,28 +1,44 @@
-Context
-=======
+Context & State
+===============
 
-Use the context block to save data into the global context to make it available for all
+Use the context block to save data into the flow context or global state to make it available for all
```
lukestanley, line 4: Not just subsequent tasks but existing ones too. Not sure how best to say this.
@the-greenman

lukestanley, line 4: Maybe good to clarify this state is in memory?

docs/workflow/blocks/context_block.rst
```diff
@@ -1,28 +1,44 @@
...
 subsequent tasks.
 
+There are two different storage domains that can be written to with this block. 
+
+**Context** is passed from block to block. When processing data in parrallel - during a batch or multiplex flow - context will be different in each branch.
+Blocks will only have access to values written by another block in the same branch. 
+
+**state** is shared by all blocks, and by all flows. When writing to or reading from State, the data is available on two different branches. 
+**state.global** is a datastore that persists for a user's entire session, no matter where they are in the site.
```
lukestanley, line 13: Can you clarify 'session'? It's not obvious if it's active memory or cookies, localstorage etc.

lukestanley, line 13: Maybe good to say that global is shared by every flow they use, so it's a bad place to put things like sensitive API session keys?

docs/workflow/blocks/context_block.rst
```diff
@@ -1,28 +1,44 @@
...
       "type": "context-save",
-      "contextKey": "savedValue"
+      "key": "saved"      
     }
 
 
 Supported properties
 --------------------
 
-- **contextKey** (string) The key of the value that will be saved into context.
```
lukestanley, line 21: We should mention it as a depreciated property so people can understand existing flows.

docs/workflow/blocks/context_block.rst
```diff
@@ -1,28 +1,44 @@
...
-  For example, if you use a key of "savedData" then the value will be available in the context as "context.savedData".
+- **key** (string) The key of the value that will be saved into context.
+  For example, if you use a key of "saved" then the value will be available in the context as "context.saved".
 
+  When context key starts with "state", values will be written to the shared state storage. State paths should
+  begin with "state.local" or "state.global". If the a key other than "local" or "global" follows "state", local will be assumed.
```
lukestanley, line 35: Maybe local by default would be safer? Is that problematic to do? @the-greenman

docs/workflow/blocks/context_block.rst
lukestanley: We should probably point that out too.

lukestanley: Someone might also expect to be able to have state persist across tabs that are both active in memory at the same time. Probably best explaining that won't work?

docs/workflow/blocks/context_block.rst
```diff
@@ -1,28 +1,44 @@
...
+  For example, if you use a key of "saved" then the value will be available in the context as "context.saved".
 
+  When context key starts with "state", values will be written to the shared state storage. State paths should
+  begin with "state.local" or "state.global". If the a key other than "local" or "global" follows "state", local will be assumed. 
+  ie. "state.key" is the equivalent of "state.local.key". (See State below)
+  
 - **valueGetter** (string) [OPTIONAL]  The JMESPath of the value to store into context. 
-  This defaults to "data" so if it is omitted it stores the whole incoming value into context.
+  This defaults to "data" so if it is omitted it stores the whole incoming value into context. 
+  Our implementation of JMESPath allows for direct manipulation of data in the JMESPath expression.
```
lukestanley, line 40: I think it's our because of our data mapping tool, which uses JMESPath. I wouldn't want people to think we have a non-standard implementation, because we don't. We just have our own set of plugins for it, used by our mapping util.

docs/workflow/blocks/mapping.rst
```diff
@@ -17,12 +24,62 @@ Default config
...
+The mapping block cannot write to state, it can only read from state. Writes are managed with the :doc:`context_block` block. 
+
+
+Local and global State
+++++++++++++++++++++++
+
+It is assumed that the state storage will mirror the site structure. Flows should generally read from *state.local*. 
+Internally *state.local* will be translated to a path to mirror the current url. 
+
+When on a the path */workflowCloud/listWorkflows*, reading from *state.local.filter* will be translated rad from  *workflowCloud.listWorkflows.filter*.
```
lukestanley, line 46: what does rad mean here?

docs/workflow/blocks/mapping.rst
```diff
@@ -17,12 +24,62 @@ Default config
...
+
+  !state.import.status 
+
+**Use this instead**
+
+.. code-block:: text
+
+  !(state.import.status) 
+
+If you want to combine a NOT statement like this with aditional logic, you need to wrap it in parentheses again.
```
lukestanley, line 74:
```suggestion
If you want to combine a NOT statement like this with additional logic, you need to wrap it in parentheses again. 
```

src/app/blocks/actions-block/actions-block.component.html
```diff
@@ -1,7 +1,7 @@
 <ng-container *ngFor="let b of buttons; let i = index">
-  <button type="button" class="actions-button" *ngIf="!b.color || b.color !== 'link'"
+  <button [attr.id]="b.id" type="button" class="actions-button" *ngIf="b.enabled && (!b.color || b.color !== 'link')"
           mat-flat-button [color]="b.color" (click)="onSubmit(b)">{{ getLabel(b) }}</button>
-  <a class="actions-button" *ngIf="b.color && b.color === 'link'"
+  <a [attr.id]="b.id" class="actions-button" *ngIf="b.color && b.color === 'link'"
```
lukestanley, line 4: Should there have been an additional e.enabled check here?
Seperately, I wonder what's going on with this color stuff.

src/app/services/shared-state.service.ts
lukestanley: Cool ok I assume it runs a tiny bit faster with a promise than a 0ms setTimeout then.

> I've attempted to use the onPush change detection strategy. And then call "this.cd.markForCheck()" when I have an update. But still get the issue.
> 
> I've tried the subscribe in multiple locations.

Interesting, can it work with some subset of blocks without defering in any way?
If this is too big a change to make to existing blocks, this could be tech debt we should accept for now, and improve on one piece at a time.

src/app/services/shared-state.service.ts
```diff
@@ -0,0 +1,94 @@
...
+ *        stateService.shared$.subscribe(state => { setTimeout(() =>{this.updateOutputDisplay()}) });
+ *       }
+ * 
+ *  The setTimeout() wrapper allows the updates to work with Angular's change detection cycle. Don't leave it out. 
+ */
+
+import { Injectable } from '@angular/core';
+import { get, set} from 'lodash-es';
+import { Subject } from 'rxjs';
+import { Location } from '@angular/common';
```
lukestanley, line 22: I didn't know Angular had their own wrapper around the location. Good to know!

src/app/services/shared-state.service.ts
```diff
@@ -0,0 +1,94 @@
...
+  ) { }
+
+  /**
+   * Returns a processed version if state that has both global and local paths. 
+   * 
+   * "global" contains the entire state tree, and is the root. 
+   * "local" contains a specific slice of the tree for the current location
+   */
+  public get state(){
+    const localPath = this.location.path().substring(1).split("/").join('.');
```
lukestanley, line 45: A comment for this would be good.
E.g: converts path format to nested path style, e.g from: `/adapterName/workflowId` to `adapterName.workflowId` to represent flow states in an addressable way.

lukestanley: Would replaceAll("/", ".") be better?

src/app/services/shared-state.service.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  /**
+   * 
+   * @param key Set a value to the internal state
+   * @param value path. 
+   * @returns Value that was set
+   */
+  setValue(key:string, value:any):any {
+    let internalKey = key;    
+    const localPath = this.location.path().substring(1).split("/").join('.');
```
lukestanley, line 61: This long line looks familiar!

src/app/services/shared-state.service.ts
```diff
@@ -0,0 +1,94 @@
...
+   * @param value path. 
+   * @returns Value that was set
+   */
+  setValue(key:string, value:any):any {
+    let internalKey = key;    
+    const localPath = this.location.path().substring(1).split("/").join('.');     
+    if (key.startsWith("global")) {
+      internalKey = key.substring(7); // remove the prefix
+    } else {
+      if (key.startsWith("local")) {
```
lukestanley, line 65: If "local" and "global" strings were constants they could be declated in once place and wouldn't need to be recreated every time they are used, and use of them could be audited more easily.

src/app/services/shared-state.service.ts
```diff
@@ -0,0 +1,94 @@
...
+    return value;
+  }
+
+  /**
+   * Returns a value from the state 
+   * @param key The path to the data to return
+   * @returns any
+   */
+  getValue(key:string):any{
+    if (!(key.startsWith("local.") || key.startsWith("global."))){
```
lukestanley, line 87: E.g this could be using constants.

docs/workflow/blocks/context_block.rst
lukestanley: Someone might think local state might be protected. Maybe good to explain it's not currently a security feature and using it as such would not work but permissions are on our roadmap and enforcing access to state may happen later?

src/app/blocks/actions-block/actions-block.component.html
the-greenman: Yeah, missing the second check. 

I've removed one of the checks for b.color. It would result in the button not being displayed if no colour was set. Now that I've added a direct way to disable a button, this should be removed.

cypress/integration/context_and_state.ts
the-greenman: I've not been able to work out how to effectively test for this grid not existing. 
My main test will always return true, whether or not the grid is displayed.

the-greenman: Ok, so you propose that, when we have tests that look for non-obvious selectors, we create a variable to clarify. Might be a good pattern.

docs/workflow/blocks/mapping.rst
the-greenman: typo.

src/app/services/shared-state.service.ts
the-greenman: Now moved to an internal function

src/app/services/shared-state.service.ts
the-greenman: Comment added. And switching to replaceAll. (Old habbits of trying to what happens if theer are duplicate characters)

src/app/services/shared-state.service.ts
the-greenman: Updated

docs/workflow/blocks/context_block.rst
the-greenman: Good point.

docs/workflow/blocks/context_block.rst
the-greenman: Additional content added to explain the protection and and persistence.

lukestanley (approved):
Without the goto sub changes and the metadata changes, I think this is probably good so I am approving it in advance.

lukestanley (approved):
👍 ✨

PR #243, opened 2021-12-10
Removing incompatible changes from angular.json

### LINKED ISSUE #239: BUG: CLI does not work
The angular CLI will no longer generate files from templates.

Example: 
`ng generate service services/shared-context `

This will produce an error

the-greenman:
This is the fix to enable the CLI to work again. It also re-enables the automatic creation of testing files. 

To test this change you can use these commands: 
 ` ng generate service services/test-service `
and 
`ng generate component blocks/test-block`

Both these should work, creating framework files. In the case of the component, a template .scss file should be created, not a .css file. 

Fixes #239

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -9,7 +9,13 @@
       "projectType": "application",
       "prefix": "app",
       "schematics": {
-      },
+        "@schematics/angular:component": {
+          "style": "scss"
+        },
+        "@schematics/angular:application": {
+          "strict": true
+        }
+      },      
       "architect": {
         "build": {
           "builder": "@angular-devkit/build-angular:browser",
--- a/angular.json
+++ b/angular.json
@@ -9,50 +9,6 @@
       "projectType": "application",
       "prefix": "app",
       "schematics": {
-        "@ngrx/schematics:component": {
-          "styleext": "scss",
-          "spec": false
-        },
-        "@ngrx/schematics:class": {
-          "spec": false
-        },
-        "@ngrx/schematics:directive": {
-          "spec": false
-        },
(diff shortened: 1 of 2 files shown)


angular.json
```diff
@@ -9,50 +9,6 @@
...
-          "spec": false
-        },
-        "@ngrx/schematics:pipe": {
-          "spec": false
-        },
-        "@ngrx/schematics:service": {
-          "spec": false
-        },
-        "@schematics/angular:component": {
-          "styleext": "scss",
```
the-greenman, line 35: So far, I've not been able to work out how to recreate this config. 

Any attempt to add this results in this error
```
Schema validation failed with the following errors:
  Data path "/schematics/@schematics~1angular:component" must NOT have additional properties(styleext).
```

the-greenman: Without this setting, the Cli will generate css files, not scss files. 



`ng generate component test`

results in these changes: 
CREATE src/app/test/test.component.html (19 bytes)
CREATE src/app/test/test.component.spec.ts (612 bytes)
CREATE src/app/test/test.component.ts (267 bytes)
CREATE src/app/test/test.component.css (0 bytes)
UPDATE src/app/app.module.ts (23424 bytes)

the-greenman: Turns out that the fix required was a change from "styleext" to "style". 

ng generate will now default to scss

lukestanley:
To test this I should run something like this below right?
`ng generate service services/new-service `?
Should anything else behave different?
@the-greenman

lukestanley (approved):
:+1:

PR #244, opened 2021-12-28
Block Metadata and an improved block editors

the-greenman:
Many of our flows contain some very complex mappings that would significantly benefit from additional descriptive info. The new state functionality may also make it easier to build flows using subroutines as it simplifies the way flows can share data. This can also increase the value of flow metadata. 

This PR adds some generic properties to all blocks, and provides some new, enhanced block editors for mapping and gosub. 

**New Properties**
These properties can now be added to any block configuration. 

- blockComment: Adds a descriptive comment to a block that will appear when the block is expanded. This will allow us to explain what is being done, and how. Very useful for more complex or unusual JMESPath. 
 The first line of the comment is used to generate a block title. 

```
{
  "type": "template",
  "blockComment":"An overview of the state"
}

```
**Upgraded Editors**

In the Mapping block, these new properties can be edited via a new, expandable "Metadata" section of the block. 

The Subroutine editor block also provides metadata editor fields, and it now also adds autocomplete dropdowns for the Adaptername and WorkflowID to make the process of working with subroutines a little easier.


docs/workflow/blocks/gosub.rst
```diff
@@ -11,12 +11,6 @@ be referenced from a form's ``uiSchema``. Or, creating dashboards
 that are made up of multiple other workflows combined with a ``multi``
 multiplex workflow block.
 
-.. code-block:: json
 
-    {
-      "type": "gosub",
-      "adapterName": "kendraio",
-      "workflowId": "myWorkflow"
-    }
```
lukestanley, line 20: Did you mean to delete this?

the-greenman: I did... because the block editor no longer lets you enter the data in that form. Like our Mapping block, it provides that lower level of structure. 

I can restore it, but put it in a different section.

lukestanley (commented):
I think comments are a great kind of metadata. But can we please have only one form of comment that a user can enter text into right away? I think that having too sections is more complicated for users than it's worth.
@the-greenman

the-greenman:
> I think comments are a great kind of metadata. But can we please have only one form of comment that a user can enter text into right away? I think that having too sections is more complicated for users than it's worth. @the-greenman

You mean to not have the expanding metadata section? Just make the form visible? I was worried that the forms could start to get too long. It also meant that the comment was displayed twice...once at the top and once below. 

Probably best to add the fields to the standard editor.

the-greenman:
@lukestanley, if you have a chance to look, I've now updated the general block editor to have title and comment on all blocks. I've also removed the comment display, and just have the form element - as it achieves the same thing.

lukestanley:
I was suggesting something like this for editing titles:

Recording 2022-01-07 at 15 10 37
This is mocked up, it doesn't save.
I moved the title editor to be edited inline exclusively.
@the-greenman

the-greenman:
I've done some investigation into in-place editing. I do think it would be handy. But, I fear that in this case, it has too many technical and UX details that need to be thought about. 

The title is currently the toggle for the expanding element, so we would need to change the behaviour. 
What happens if the standard title is long? Where does the box go?

the-greenman:
But did you actually implement that as in-place editing?

lukestanley:
I did.

lukestanley:
Long titles can wrap, which works nicely.
I was able to disable opening up the panel when editing the title.
Maybe I should put it in a branch for you to try.
Another idea is to have an edit button which would leave more room for
clicking on the title.

the-greenman:
If you can share your updates in a branch it would be really useful!

lukestanley:
> If you can share your updates in a branch it would be really useful!

Here: https://github.com/kendraio/kendraio-app/tree/blockMetadataInlineExperiment
It doesn't actually save the edits but it does show what it'd be like.
@the-greenman

the-greenman:
@dahacouk Do you have some thoughts on this kind of in-place editing interface? 

It's not quite as easy to implement as I expected, and may contain a few gotchas, but it could be valuable. The extra work to implement this kind of interface could also be left for the future if we want to try a simpler version first.

src/app/components/blocks-editor/blocks-editor.component.ts
```diff
@@ -16,21 +16,40 @@ BLOCK_TYPES.forEach(element => _blockTypes[element.type]=element);
...
 
-  ngOnInit() {
+  ngOnInit() {    
+  }
+
+  ngOnChanges(){
+    this.generateTitles();
+  }
+
+  generateTitles(){
```
the-greenman, line 39: Generating block titles from comments.

src/app/components/blocks-editor/blocks-editor.component.ts
```diff
@@ -16,21 +16,40 @@ BLOCK_TYPES.forEach(element => _blockTypes[element.type]=element);
...
+    this.blocks.forEach((block, i) => {            
+      const maxLength = 30       
+      let comment  = block.blockComment ? block.blockComment.split("\n")[0]: '';
+      // if the length is longer than max, truncate at word boundary
+      if (comment.length > maxLength) {
+        comment = comment.substr(0, comment.lastIndexOf(" ", maxLength)) + "...";
+      }      
+      this.blockTitles[i] = comment;
+    });
```
lukestanley, line 48: Why not use it directly but style it with CSS?
https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow
This gives more control to the presentation layer that knows what the real limits are.
@the-greenman

lukestanley: 97% of browsers support it:
https://caniuse.com/text-overflow

the-greenman: Suppose that would be fine.

the-greenman: Now implemented with text-overflow. 
Still only getting the first line to allow for control of what appears.

lukestanley (approved):

Issue #246, opened 2022-01-07
Fix flakey Cypress tests


lukestanley:
- Cypress was sometimes failing to load the initial web app webpage from Vercel.

- In a mapping block E2E test,
[The mapping block fails to disappear, intermittently](https://github.com/kendraio/kendraio-app/runs/4737899699?check_suite_focus=true#step:3:455) before the timeout.
This fixes it by improving the timeout of the assertion.

> "should make new workflow with mapping block:
       AssertionError: Timed out retrying after 4000ms: Expected not to find content: 'Mapping' within the element: <app-workflow-sidenav.ng-tns-c342-3> but continuously found it."

the-greenman:
Not seeing these anymore

PR #252, opened 2022-01-25
CI reliability fixes: expands timeouts duration, disables unused video capture #246

lukestanley:
This PR fixes observed Cypress CI failures by:
- increasing the page load timeout, 
- increases the default Cypress command timeouts.
- disabling videos.


the-greenman (approved):
My test for this was: will it pass the tests that are currently failing. Looks like it worked first time, while the existing config failed three times in a row.

PR #253, opened 2022-01-26
Deactivate guard

the-greenman:
This is an implementation of a popup message to warn a user that some changes to the workflow have not been saved.

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -108,7 +108,7 @@
                 },
                 "fonts": false
               },
-              "sourceMap": false,
+              "sourceMap": true,
               "namedChunks": false,
               "extractLicenses": false,
               "vendorChunk": false,
--- a/src/app/services/dirty-dialog.service.spec.ts
+++ b/src/app/services/dirty-dialog.service.spec.ts
@@ -1,0 +1,16 @@
+import { TestBed } from '@angular/core/testing';
+
+import { DirtyDialogService } from './dirty-dialog.service';
+
+describe('DirtyDialogService', () => {
+  let service: DirtyDialogService;
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({});
+    service = TestBed.inject(DirtyDialogService);
+  });
+
+  it('should be created', () => {
+    expect(service).toBeTruthy();
+  });
+});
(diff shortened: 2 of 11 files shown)


the-greenman:
https://user-images.githubusercontent.com/166254/151370624-aa931998-a853-4de3-9b2a-8b75a7778045.mp4

the-greenman:
Now with a working and tested guard to prevent loss of changes

angular.json
```diff
@@ -108,7 +108,7 @@
                 },
                 "fonts": false
               },
-              "sourceMap": false,
+              "sourceMap": true,
```
the-greenman, line 111: This setting disabled the ability to debug! 
Can move into a separate branch if necessary

src/app/services/dirty-dialog.service.ts
```diff
@@ -0,0 +1,22 @@
+import { Injectable } from '@angular/core';
+import { Observable, of } from 'rxjs';
+
+/**
+ * Async modal dialog service
+ * DialogService makes this app easier to test by faking this service.
```
lukestanley, line 6: Do you mean "DirtyDialogService"?
@the-greenman

src/app/services/dirty-dialog.service.spec.ts
```diff
@@ -0,0 +1,16 @@
...
+describe('DirtyDialogService', () => {
+  let service: DirtyDialogService;
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({});
+    service = TestBed.inject(DirtyDialogService);
+  });
+
+  it('should be created', () => {
+    expect(service).toBeTruthy();
```
lukestanley, line 14: Is this useful? @the-greenman

the-greenman: I'm wondering about these auto generated stubs. They do provide an essential bit of test coverage. They do actually make sure that the object gets created. 

In this case I don't actually know how to test the current functionality of this service outside of Cypress, but functionality may evolve. 

Getting back to the Angular default standard of always having a spec test for a file may be a good objective.

src/app/services/dirty-dialog.service.ts
the-greenman: Was an old comment anyway. Removed

lukestanley (approved):

PR #256, opened 2022-02-04
File input extended to support state

the-greenman:
This PR updates the file input block to be state aware, and adds "EnabledGetter" functionality to allow the button to be revealed or hidden based on state.

### DIFF
--- a/cypress/fixtures/develModeBlocks.json
+++ b/cypress/fixtures/develModeBlocks.json
@@ -1,0 +1,78 @@
+{
+    "title": "Workflow",
+    "blocks": [
+      {
+            "type": "context-save",
+            "valueGetter": "`false`",
+            "contextKey": "state.flags.devel"
+        },
+        {
+            "type": "context-save",
+            "valueGetter": "`true`",
+            "contextKey": "state.flags.verbose"
+        },
+      {
+        "type": "template",
+        "template": "{{#unless state.flags.devel}}Not in Develop mode{{/unless}}{{#if state.flags.devel}}Develop mode{{/if}}",
+        "blockComment": ""
+      },
+      {
+        "type": "debug",
+        "open": 1,
+        "showData": true,
+        "showContext": false,
+        "showState": false,
+        "enabledGetter": "state.flags.devel",
+        "blockComment": ""
+      },
+      {
+        "type": "actions",
+        "buttons": [
+          {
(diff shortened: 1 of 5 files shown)


cypress/fixtures/develModeBlocks.json
```diff
@@ -0,0 +1,78 @@
+{
```
the-greenman, line 1: I've created this fixture to simplify future testing of these features. 
It defines a standard set of blocks that should be state aware, and removes the need to define blocks in each test. 


Older tests have not yet been refactored to use it.

src/app/blocks/file-input-block/file-input-block.component.ts
```diff
@@ -1,6 +1,8 @@
 import {Component} from '@angular/core';
```
the-greenman, line 1: There are a fair number of changes required to support this functionality. 
I have been considering adding it to the base class, or creating a new extension of the base class to make it easier to implement. 
Just not this time.

cypress/fixtures/develModeBlocks.json
lukestanley: Couldn't this sort of comment could go in the JSON itself, and maybe the filename could be more obvious?

the-greenman: Done. Comment and file rename

cypress/integration/context_and_state.ts
```diff
@@ -133,5 +133,14 @@ it('should allow the disabling of the debug block', () => {
...
-
-});
+it('should allow the disabling of the file input block ', () => {
+  cy.intercept('GET', 'https://app.kendra.io/api/TESTING/develmodeblocks', {
+    fixture: 'develModeBlocks.json' }
+    ).as('workflow.json');
+    cy.visit('/TESTING/develmodeblocks');
+    cy.contains('Devel Import').should('not.exist');
+    cy.contains("Enable Devel Mode").click();
+    cy.contains('Devel Import').should('exist');
```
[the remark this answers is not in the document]
the-greenman: Yeah, good point. I've changed the label on the button to be more verbose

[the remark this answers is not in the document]
the-greenman: I've updated the labels here

lukestanley (approved):

PR #258, opened 2022-02-09
Cypress test reliability improvements

lukestanley:
- Retries are enabled
- Font loading requests, and a few API requests are intercepted.
- Time waiting for a page to load has been reduced, in favour of retries.
- Code linting / formatting improvements


the-greenman (approved):
Looking forward to seeing how this improves things. 
We should look at a standard set of setup steps that we can use in the future, perhaps turning these intercepts into a shared command

PR #264, opened 2022-02-21
Check for valid data before loading a schema

the-greenman:
This is a tiny fix to check that we have a valid schema Id in place before we attempt to load it. 

I've not got a test written for this one. This is essentially a bug fix, so I'm not creating a test for the existence of the bug.

### DIFF
--- a/src/app/blocks/load-schema-block/load-schema-block.component.ts
+++ b/src/app/blocks/load-schema-block/load-schema-block.component.ts
@@ -32,15 +32,16 @@ export class LoadSchemaBlockComponent  extends BaseBlockComponent {
     const baseSchema = (!!this.schemaGetter)
       ? mappingUtility({ context: this.context, data: this.model }, this.schemaGetter)
       : this.schema;
-
-    const schemaDefinitions = {};
-    schemaDefinitions[baseSchema] = await this.resolveSchema(schemaDefinitions, baseSchema);
-    const jsonSchema = {
-      definitions: schemaDefinitions,
-      '$ref': `#/definitions/${baseSchema}`
-    };
-    // TODO: some fields may need uiSchema (eg widget overrides)
-    this.output.emit({ jsonSchema, uiSchema: {} });
+    if((typeof baseSchema === 'string') && baseSchema.length > 0) {
+      let schemaDefinitions = {};
+      schemaDefinitions[baseSchema] = await this.resolveSchema(schemaDefinitions, baseSchema);
+      const jsonSchema = {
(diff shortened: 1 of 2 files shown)


src/app/blocks/load-schema-block/load-schema-block.component.ts
```diff
@@ -32,15 +32,16 @@ export class LoadSchemaBlockComponent  extends BaseBlockComponent {
...
-
-    const schemaDefinitions = {};
-    schemaDefinitions[baseSchema] = await this.resolveSchema(schemaDefinitions, baseSchema);
-    const jsonSchema = {
-      definitions: schemaDefinitions,
-      '$ref': `#/definitions/${baseSchema}`
-    };
-    // TODO: some fields may need uiSchema (eg widget overrides)
-    this.output.emit({ jsonSchema, uiSchema: {} });
+    if((typeof baseSchema === 'string') && baseSchema.length > 0) {
```
lukestanley, line 35: This could be more symetrical, e.g:
```suggestion
    if((typeof baseSchema === 'string') && (baseSchema.length > 0)) {
```

src/app/blocks/rename-fields-block/rename-fields-block.component.ts
```diff
@@ -41,7 +43,15 @@ export class RenameFieldsBlockComponent  extends BaseBlockComponent {
   mapItem(item, fieldMapping) {
     const output = {};
     (fieldMapping || []).forEach(({ source, target }) => {
-      output[target] = item[source];
+      //set default mapping direction to sourceToTarget
+      let from = source;
+      let to = target;
+      if (this.direction == 'targetToSource') {
```
lukestanley, line 49: This seems to be a new feature, not just a bug fix. This could be explained better.

lukestanley (changes requested):
The existing functionality is not well tested, nor are the changes, e.g: targetToSource or error checking.

src/app/blocks/load-schema-block/load-schema-block.component.ts
lukestanley: Or maybe just the outer brackets.

the-greenman: done.

the-greenman:
I've removed the accidentally added new code to this.  We need a better grip on how we are managing schemas in general before creating tests for this block. 

This is now nothing more than a small big fix to remove an error in mapping code.

lukestanley (approved):

PR #271, opened 2022-03-25
Allow complete adapters to be "compiled" and accessed as data

the-greenman:
This PR refines the adapter install service to make it's exportable adapter data available internally, and then extends the adapter-info block to allow it to access this data.


src/app/services/adapter-install.service.ts
```diff
@@ -95,7 +95,14 @@ export class AdapterInstallService {
...
     const exportData = {...adapterConfig, workflow, database, forms, attachments};
-    this.downloadData(exportData, adapterName);
+    return exportData;
+  }
+
+  async exportAdapter(adapterConfig) {
+    const {adapterName} = adapterConfig;
+    this.compileAdapter(adapterConfig).then(exportData => {
+      this.downloadData(exportData, adapterName);
+    });
```
the-greenman, line 105: I have split the exportAdapter function in two, with the meat moved to "compileAdapter".

src/app/blocks/adapter-info-block/adapter-info-block.component.ts
```diff
@@ -41,10 +47,19 @@ export class AdapterInfoBlockComponent extends BaseBlockComponent {
...
+        if (this.compileAdapter) {
+          this.adapterInstallService.compileAdapter(adapterData).then((compiledAdapter) => {
+            console.log(compiledAdapter);
+            this.isLoading = false;
+            this.output.emit(compiledAdapter);
+          });
+        } else {
+          this.isLoading = false;
+          this.output.emit(adapterData);
+        }
```
the-greenman, line 62: If the compileAdapter flag is set, we use the AdapterInstallService to gather up all the adapter components, and return them all in a format compatible with export

src/app/blocks/graphql-block/graphql-block.component.ts
```diff
@@ -3,6 +3,7 @@ import {BaseBlockComponent} from '../base-block/base-block.component';
 import {HttpClient} from '@angular/common/http';
 import { get, isObject } from 'lodash';
 import {mappingUtility} from '../mapping-block/mapping-util';
+import { SharedStateService } from 'src/app/services/shared-state.service';
```
the-greenman, line 6: GraphQl gets a small update to allow it to access state. This makes it possible to access shared git config data.

src/app/services/core-event-handlers.service.ts
```diff
@@ -89,12 +90,18 @@ export class CoreEventHandlersService {
     this.formSubmit.actions$
       .pipe(
       )
-      .subscribe(({ action }) => {
-        // console.log(`process action ${action}`);
+      .subscribe(({action, payload}) => {
```
the-greenman, line 93: This change is required in order to test the export adapter functionality. 
The process of installing an adapter via the standard, core, flows felt impossible to achieve. Because the process relied on local storage and local index db, most of the steps involved in setting up the tests did not reliably work.

lukestanley:
Would 'package' instead of 'compiled' make more sense? @the-greenman

src/app/blocks/graphql-block/graphql-block.component.ts
```diff
@@ -49,12 +50,13 @@ export class GraphqlBlockComponent extends BaseBlockComponent {
       payload = {
         query: this.query,
         variables: Object.keys(this.variableGetters).reduce((a, key) => {
-          a[key] = mappingUtility({ data, context: this.context }, this.variableGetters[key]);
+          a[key] = mappingUtility({ data, context: this.context, state:this.stateService.state }, this.variableGetters[key]);
           return a;
         }, {})
       };
+      console.log('payload', payload);
```
lukestanley, line 57: Should the log stay here? @the-greenman

src/app/services/core-event-handlers.service.ts
```diff
@@ -89,12 +90,18 @@ export class CoreEventHandlersService {
...
-        // console.log(`process action ${action}`);
+      .subscribe(({action, payload}) => {         
         if (action === 'refreshWorkflow') {
           this.workflow.refresh();
           return;
         }
+        /**
+         * Allow adapters to be installed by non-core flows?
+         */
+        if (action === 'installAdapter'){
```
lukestanley, line 101: Should we have a flag for this? E.g a debug flag?
Or maybe there is a good way to stub / substitute a flag or the core/non-core status?
@the-greenman

the-greenman:
> Would 'package' instead of 'compiled' make more sense? @the-greenman

Yeah. Better name. I've updated.

src/app/services/core-event-handlers.service.ts
the-greenman: This has now been removed. New global settings flag introduced.

src/app/services/core-event-handlers.service.ts
```diff
@@ -21,11 +22,13 @@ export class CoreEventHandlersService {
     private readonly workflow: WorkflowService,
     private readonly localData: LocalDatabaseService,
     private readonly adapterInstall: AdapterInstallService,
-    private readonly translate: TranslateService
+    private readonly translate: TranslateService,
+    private readonly settings: AppSettingsService
   ) {
+    const expose = this.settings.get('exposeCoreActions', false);
```
the-greenman, line 28: The main settings page now has an option to "Expose core actions". If this is enabled, we allow workflows to access these.

src/app/blocks/graphql-block/graphql-block.component.ts
the-greenman: Gone now

src/app/services/core-event-handlers.service.ts
the-greenman: I've now added a new settings flag to the core settings form.

cypress.json
```diff
@@ -17,5 +17,6 @@
   "watchForFileChanges":false,
   "projectId": "ebzwp1",
   "pageLoadTimeout": 200000,
-  "defaultCommandTimeout": 30000
+  "defaultCommandTimeout": 30000,
+  "experimentalStudio" : true
```
lukestanley, line 21: I guess we can leave this in, so long as CI still works well. @the-greenman

cypress/integration/adapters.ts
```diff
@@ -0,0 +1,56 @@
...
+          "type":"adapter-info",
+          "adapterName":"bloomen",
+          "packageAdapter":true
+        },
+        {
+          "type":"debug",
+          "open":4
+        },
+      ]);
+     cy.contains("attachments").should('exist');
```
lukestanley, line 51: Would prefer comments saying what just happened. E.g what is an attachment? @the-greenman

cypress/integration/spec.ts
```diff
@@ -9,7 +9,12 @@ describe('workspace-project App', () => {
     cy.intercept('GET', 'https://kendraio.github.io/kendraio-adapter/config.json', {
       fixture: 'adapterConfig.json'
     }
-    ).as('adapterConfig.json');
+    ).as('adapterConfig.json');    
+    // Prevent external network request for adapter config
+    cy.intercept('GET', 'https://app.kendra.io/api/workflowCloud/listWorkflows', {
```
lukestanley, line 14: :+1:

src/app/services/core-event-handlers.service.ts
lukestanley: Wouldn't it be better if this comment was in the code? @the-greenman

src/app/services/core-event-handlers.service.ts
```diff
@@ -21,11 +22,13 @@ export class CoreEventHandlersService {
...
     this.formSubmit.actions$
-      .pipe(
-        filter(({ form }) => form === 'core')
+      .pipe(        
+        filter(({ form }) => (form === 'core') || expose)
```
lukestanley, line 31: I know what this does right now but can it be made explicit with some comments or such? @the-greenman

lukestanley (approved):
I am mostly happy with the code, but some comments would help make it more obvious and maintainable, so I am approving in advance.

src/app/services/core-event-handlers.service.ts
the-greenman: Comment added

cypress.json
the-greenman: Yeah. My thoughts. Can't ignore the file.

PR #275, opened 2022-04-20
Adapter upgrades

the-greenman:
This PR contains all the recent updates for adapters, and the sponsor info.

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -17,5 +17,6 @@
   "watchForFileChanges":false,
   "projectId": "ebzwp1",
   "pageLoadTimeout": 200000,
-  "defaultCommandTimeout": 30000
+  "defaultCommandTimeout": 30000,
+  "experimentalStudio" : true
 }
--- a/cypress.json
+++ b/cypress.json
@@ -2,6 +2,7 @@
   "integrationFolder": "cypress/integration",
   "supportFile": false,
   "videosFolder": "cypress/videos",
+  "video":false,
   "screenshotsFolder": "cypress/screenshots",
   "pluginsFile": false,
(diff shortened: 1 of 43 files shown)


cypress.json
```diff
@@ -17,5 +17,6 @@
   "watchForFileChanges":false,
   "projectId": "ebzwp1",
   "pageLoadTimeout": 200000,
-  "defaultCommandTimeout": 30000
+  "defaultCommandTimeout": 30000,
+  "experimentalStudio" : true
```
lukestanley, line 21: Have you noticed if the speed on CI is impacted by this? @the-greenman

cypress/integration/adapters.ts
```diff
@@ -0,0 +1,61 @@
...
+      fixture: 'bloomen.json'
+    }
+    ).as('bloomen.json');
+
+  });
+
+  it('Export a packaged adapter with attachments', () => {
+    // first go to settings and allow core actions in flows
+        cy.visit('/core/settings');
+        cy.get('#formly_16_boolean_exposeCoreActions_5 .mat-checkbox-inner-container').click();
```
lukestanley, line 24: I wonder how well this selector will hold.

lukestanley (approved):

cypress.json
the-greenman: I didn't see any change. It almost looked faster.

cypress/integration/adapters.ts
the-greenman: Took a while to find a selector that eventually worked. I failed to get it to work with a "contains" statement, as there are a number of nested elements, that seemed to get in the way. I hope that the ID of the field will be fairly static.

PR #277, opened 2022-04-26
Test fix

the-greenman:
Attempting to change tests to be more reliable.


the-greenman:
General recommendations for less flaky tests is to chain commands together. 
So far, it seems like it's working a bit better

lukestanley:
I think I get what's going on a bit better now. Your code is in the right direction mostly. The reason why it works more reliably is because the contains("someAdapterB") assertion check is repeated, along with the previous steps in the chain until the assertion passes (if it can pass within a certain threshold).

PR #287, opened 2022-06-06
286 dependabot updates

the-greenman:
This is a testing branch for integrating the security updates from dependabot.


the-greenman:
The Dexie changes here seem fairly significant, but, I've not seen any issues. Have tested a number of flows that do localdb stuff.

lukestanley (approved):

PR #290, opened 2022-06-13
New mermaid block

the-greenman:
A new block type to implement mermaid. This will allow for the creation of simple graphs and visualisations.


angular.json
```diff
@@ -117,7 +117,7 @@
                 {
                   "type": "initial",
                   "maximumWarning": "10mb",
-                  "maximumError": "23mb"
```
the-greenman, line 120: We've only bumped mermaid one point release. The size increase has been cumulative. We should investigate soon.

docs/workflow/blocks/mapping.rst
```diff
@@ -155,6 +155,29 @@ In this example, we merge two object arrays, and look up a unique key from a lis
   find($.data.licensor, 'name', "LICENSOR NAME").id,   
 
 
+Extracting content from different levels in a nested array
```
the-greenman, line 158: Some notes from working on the schema extraction

lukestanley (approved):

PR #292, opened 2022-06-13
Getting deployment working on Firebase

the-greenman:
Firebase deployments have a more strict approach to dependencies. Recent package updates have caused firebase deployments to fail, and now we are not able to push to production. 

The cause is a depreciated package: @angular-devkit/build-ng-packagr

Upgrading devkit core seems to be the recommend step.


package.json
```diff
@@ -99,10 +97,10 @@
     "zone.js": "~0.11.4"
   },
   "devDependencies": {
-    "@angular-devkit/build-angular": "^13.0.1",
-    "@angular-devkit/build-ng-packagr": "^0.1100.0-next.5",
+    "@angular-devkit/build-angular": "^13.3.7",
+    "@angular-devkit/core": "^14.0.1",
```
the-greenman, line 101: Version 14 of devkit core seems fine. I didn't want to pin to anything older than necessary.

PR #303, opened 2022-06-28
Auth0 Callback page now has default behaviour

the-greenman:
This update provides a simple additional behaviour for the auth0 callback page. If no authentication data is passed in the url, then the user is forwarded directly to the user page. A user should never be left on this page, and since this is an authentication callback, we can safely send users to the /user page. 

The Auth0 login process for Youtube seems to cause the callback page to be reloaded after authentication is complete. The second call is missing the authentication payload, and without any default action, we get stuck. 

I would like to find the source of the second redirect, but it has so far taken enough time, and may not be worth further investment.

### DIFF
--- a/cypress/integration/authentication.ts
+++ b/cypress/integration/authentication.ts
@@ -1,0 +1,11 @@
+import { loadFlowCode } from '../support/helper';
+// tslint:disable: quotemark
+/// <reference types="Cypress" />
+
+describe('Old Account page', () => {
+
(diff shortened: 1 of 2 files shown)


lukestanley (approved):
The if statements could already be more simple, and I know you didn't write it, you've added to them - you've also fixed a bug and added a test so I'll approve this, but we should probably have a ticket to reduce the complexity of the auth service branch conditions.
https://grugbrain.dev

PR #309, opened 2022-07-14
Codespaces config

the-greenman:
This is an initial setup to get codespaces working. 

Minimal devcontainer definition and readme file updates.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,5 +1,7 @@
 {
     "name":"kendraio-app",
-    "forwardPorts": [4200]
+    "forwardPorts": [4200],
+
+    "image": "mcr.microsoft.com/vscode/devcontainers/typescript-node:0-16"
 
 }
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,5 @@
+{
+    "name":"kendraio-app",
+    "forwardPorts": [4200]
+
+}
--- a/angular.json
+++ b/angular.json
@@ -132,7 +132,8 @@
         "serve": {
           "builder": "@angular-devkit/build-angular:dev-server",
           "options": {
-            "browserTarget": "kendraio-app:build"
+            "browserTarget": "kendraio-app:build",
+            "host": "0.0.0.0"
           },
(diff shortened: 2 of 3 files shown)


lukestanley:
"A devcontainer.json file in your project tells Visual Studio Code (and other services and tools that support the format) how to access (or create) a development container with a well-defined tool and runtime stack. It's currently supported by the [Remote - Containers](https://aka.ms/vscode-remote/download/containers) extension and [GitHub Codespaces](https://github.com/features/codespaces)."

https://code.visualstudio.com/docs/remote/devcontainerjson-reference
While mentioning the Microsoft / GitHub product or service it works with right now is useful, I think that generalising it too would be good. It's good that this will support more than just Codespaces, VS Code and maybe VS Codium, I wouldn't be surprised if JetBrains have support now or soon too.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,7 @@
+{
+    "name":"kendraio-app",
+    "forwardPorts": [4200],
+
+    "image": "mcr.microsoft.com/vscode/devcontainers/typescript-node:0-16"
```
lukestanley, line 5: Why this specific image? @the-greenman

the-greenman: I thought it was a good idea to pick a particular image that was optimised for our kind of work, rather than a general one. This is a step towards looking at our own docker config that would be optimised for our work.

angular.json
```diff
@@ -132,7 +132,9 @@
         "serve": {
           "builder": "@angular-devkit/build-angular:dev-server",
           "options": {
-            "browserTarget": "kendraio-app:build"
+            "browserTarget": "kendraio-app:build",
+            "host": "0.0.0.0",
+            "disableHostCheck": true
```
the-greenman, line 137: This is another change that is required to work with remote containers on dynamic urls. 

I've not been able to find much discussion around these changes. They all introduce potential issues. 
However, most of the issues result in access to source code, which is not a problem for us. 

They also could potentially reveal secrets stored on a server... but since we store all data in the client, this does not impact us either. 

So, I don't think we need to worry. 


We will however get warnings when we start a dev server.

PR #311, opened 2022-07-26
Load schema tests and documentation

lukestanley:
- Documents  load-schema block functionality better 

- Adds some Angular tests for key new load-schema block functionality.


the-greenman (approved):
Looking good. No suggestions from me

PR #313, opened 2022-08-10
Adds record template to schema converter and DB block upsert method

lukestanley:
# Overview:

- Introduces an "upsert" method to the DB block with documentation and tests.

- Adds a converter for the load-schema block, with documentation and unit tests.

Currently the converter uses the CMS related config key name: "blockTypeDefaults",
This could be replaced with something like "defaultInjections". 
It could allow multiple user specified target keys to search for applying the injection of default data to, as well.

Below is pasted from the Sphinx documentation for the converter:

# Convert template records into form schemas:

Makes form schemas from existing JSON data "templates". Rather than records being fixed in stone with no way to add new properties, we enable making forms dynamically that build upon existing JSON data.

Using existing JSON records as a template, we can create new forms for making records of data. How? By using transformations that users can configure to, adding new fields, and rename keys. To make descendants of templates uniquely identifiable, we can also generate new UUIDs. By transforming existing JSON record objects using user defined templates, data is changed into a form templates using configurable rules.

## Background context:

Users can make schemas using the schema builder, and make records for the schemas. The records can themselves be templates for other forms. Users can make records with the Form block and save them as records for use as templates. Example use-case: a Content Management System (CMS) can have templates for different blocks of content, text, images, videos etc. Using this, the CMS can then use the templates to create new content. The load-schema block can allow adding new JSON properties to these records. We convert template objects so further information can be added to the form. We make new JSON schemas for the Form block to use to fill extra details in,

## Terminology:

We use the term "template" to refer to a JSON object that is used as a basis for making a form. The term "form" is used to refer to a JSON object that is used to make a record using the form block.

## Currently focused on the CMS use-case:

Currently, the implementation is focused on the CMS (Content Management System) use-case.

Given JSON record, created from an existing form like this:

```json
{
    "websitemodel": {
        "blockmodel": [
            {
                "uuid": "c5bfac02-f0b6-4c31-8fdf-18de02667ee9",
                "src": "https://placekitten.com/200/300"
            }
        ]
    }
}
```

The load-schema block can be loaded and configured like this:

```json
{
  "block": "load-schema",
  "templateToSchema": {
    "blockTypeDefaults": {
      "c5bfac02-f0b6-4c31-8fdf-18de02667ee9": {
        "type": "object",
        "properties": {
          "src": {
            "type": "string",
            "title": "Image URL"
          }
        }
      }
    },
    "convertables": {
      "websitemodel": "website-content",
      "blockmodel": "block-content"
    }
  }
}
```

It will use that to produce a new JSON schema for the form block to use, something like this:

```json
{
    "type": "object",
    "properties": {
        "website-content": {
            "type": "object",
            "properties": {
                "block-content": {
                    "type": "object",
                    "properties": {
                        "b74e30a7-a052-4331-86eb-ba2eb7a31ce4": {
                            "type": "object",
                            "properties": {
                                "src": {
                                    "type": "string",
                                    "title": "Image URL"
                                }
                            }
                        }
                    }
                }
            }
        }
    }
}
```

It uses the existing loaded JSON record data as context, and output JSON Schema as config for use by the Form block (to allow editing).

It can also be saved as a JSON schema, to later load and make multiple "child" documents.

Note the UUIDs in the JSON Schema are generated automatically for the "block-content" property items.


src/app/blocks/db-block/db-block.component.ts
```diff
@@ -122,6 +122,43 @@ export class DbBlockComponent extends BaseBlockComponent {
         }.bind(this));
         return;
       }
+      case 'upsert': {
```
the-greenman, line 125: Would it not be worth implementing the "Put" function for Dexie.js in the localDatabaseService? 
That handles all the update or replace logic too.

docs/workflow/blocks/load_schema.rst
```diff
@@ -89,8 +89,8 @@ If the above is loaded in (using a database / DB block, like the "editItem" flow
     }
 
 
-Reference transclusion
-----------------------
+Referencing / transcluding other schemas (via array / object config)
```
the-greenman, line 92: I suspect some people will be put off by the "transclude" word. It's not one commonly used. We could benefit from a more general description.

"Referencing structures and data from other schemas"

src/app/blocks/load-schema-block/load-schema-block.component.ts
```diff
@@ -4,6 +4,157 @@ import { get, has, clone } from 'lodash-es';
...
+      }
+    }
+    return schema
+  }
+  console.log("returning test")
+  //return blockTypeDefaults["c5bfac02-f0b6-4c31-8fdf-18de02667ee9"];
+  return convertToValidJSONSchema(arrayOfUUIDsToDict(applyBlockTypeDefaults(convertKeys(template))))
+}
+
+
```
the-greenman, line 157: This probably needs some introduction or overview. It's a chunk of new tech that may be core to future work. I'd like to little intro to what is going on to that I understand the higher level concepts, before needing to get into the details of the code.

src/app/blocks/db-block/db-block.component.ts
lukestanley: Yes I started that and put it in a branch to investigate another time: https://github.com/kendraio/kendraio-app/tree/local-database-put-method

docs/workflow/blocks/load_schema.rst
lukestanley: Yes, I've made a few of the titles and explanations more clear with less jargon right away, thank you.

src/app/blocks/load-schema-block/load-schema-block.component.ts
lukestanley: You should find it better commented now.

lukestanley:
I haven't added any Read-the-docs documentation yet but I think that this feature will be in flux a lot, so I'm not sure it's worth it yet. What do you think?

lukestanley:
I added some Sphinx documentation for how to use the feature.

lukestanley:
I only added documentation but now the previously passing Cypress tests are failing, complaining about an invalid baseUrl that I cannot find in the source.

lukestanley:
I reran the tests and it is happy again now. Weird way to be flakey though.

the-greenman (approved):
My main comment on the comments is that I prefer JSDoc style comments at the top of each key function. These provide the overview of the code, and most IDEs are aware of them, and can then provide pop-up docs and autocomplete. 

I don't think this should prevent me approving this, but it's something we should think about for a style guide.

lukestanley:
It is now using JSDoc style comments in the converter, and the tests seem less flakey! :)

PR #314, opened 2022-08-30
Container Based Development and Testing

the-greenman:
This is an update to the development container setup, and the main package file to allow for development and testing in containers. 

These changes include: 
* Devcontainer config
* xvfb-maybe config for testing commands


package.json
```diff
@@ -97,7 +97,7 @@
     "zone.js": "~0.11.4"
   },
   "devDependencies": {
-    "@angular-devkit/build-angular": "^13.3.7",
+    "@angular-devkit/build-angular": "^13.3.9",
```
the-greenman, line 100: Minor upgrades to test tools. These are all working. I feel it's work keeping the updates happening.

the-greenman:
Testing now works "out of the box" on github devcontainers.

the-greenman:
But our main github testing process is now generally flakey

the-greenman:
I've never been able to get one of the cypress tests to pass on this branch. 
I've now downgraded the cypress package again to see if that is an issue.

the-greenman:
> I've never been able to get one of the cypress tests to pass on this branch. I've now downgraded the cypress package again to see if that is an issue.

Hmm. Looks cypress and electron updates may require some additional investigation. I'll create an issue and see if we can isolate the problem.

lukestanley (approved):

PR #322, opened 2022-10-11
Cypress10

the-greenman:
An update to Cypress 10, with all tests renamed and config structure updated.


package.json
```diff
@@ -42,6 +42,7 @@
     "@daz.is/jmespath": "^1.3.1",
     "@fortawesome/fontawesome-free": "^5.15.4",
     "@ng-select/ng-select": "^7.4.0",
+    "@ngtools/webpack": "^14.2.5",
```
the-greenman, line 45: Webpack required an upgrade to work

cypress.config.ts
```diff
@@ -0,0 +1,30 @@
+import { defineConfig } from 'cypress'
```
the-greenman, line 1: Config is now code, not JSON file

lukestanley (approved):
:+1:

PR #323, opened 2022-10-12
State reactive mapping block

lukestanley:
Allows the mapping block to subscribe to state update events (unless disabled by config - see tests)
Throttles when loop is likely.


lukestanley:
Thanks. It can loop forever though, just, very slowly. :)

PR #326, opened 2022-10-19
Disabled buttons rather than hidden

the-greenman:
An updated version of the actions block that disables buttons rather than hiding them.


the-greenman:
Ahhh... programmer bliss. Tests failed because my changes broke the feature!! Well, they changed the feature and the tests detected the change.

lukestanley:
@gsambrotta can you review this please and check it works as intended?

PR #327, opened 2022-10-20
Angular13

the-greenman:
This is an update to Angular 13. All dependencies are now sorted.


cypress/e2e/adapters.ts
```diff
@@ -32,11 +33,24 @@ describe('Adapter install', () => {
...
+                "action": "installAdapter"
+              }]
+          }]
       }
     ]);
-    cy.wait(1000); // give the adapter time to install.
+    
+    let currentConfig=localStorage.getItem('core.variables.settings');
+    cy.log("Local:").log(currentConfig);
+    cy.contains('Import Adapter').click().wait(['@bloomen']).wait(2000); // make sure that the data is downloaded and give the adapter time to install.
```
the-greenman, line 53: If the settings are not in place, this call never happens, because it is blocked in the code.

angular.json
```diff
@@ -223,6 +223,7 @@
   },
   "defaultProject": "kendraio-app",
   "cli": {
-    "defaultCollection": "@ngrx/schematics"
+    "defaultCollection": "@ngrx/schematics",
+    "analytics": false
```
the-greenman, line 227: Don't send data back to Angular!

PR #328, opened 2022-10-20
Share schemas and records via URL (local metadata db table)

lukestanley:
When pressing share in a flow, the user must now choose to share the flow or the database.
A test ensures that:
- the metadata database table state can be loaded just with the URL
- a flow can be loaded from a URL
- the same database state is exactly reproduced when saving.


the-greenman (approved):
At some point, we'll need to find a way to not be sharing/wiping the whole DB. 
For now, this works... but lets not forget!!

PR #329, opened 2022-10-24
Angular14

the-greenman:
Upgraded to Angular 14. 
Changes in code required for this.


the-greenman:
Karma test runner failing. May need a change in syntax.

lukestanley:
Locally I was able to run the Angular unit tests via `npx ng test` (did not use devcontainer).

the-greenman:
Yeah. I also have no trouble running tests locally. This issue has made me think that we probably have way too many different environments. Github tests use one, Vercel uses one, production a different one.. and then locally too.

lukestanley:
Despite trying for a bit, I was not able to reproduce the below issue locally by running the same base Docker image.
```
  Option '--' has been specified multiple times. The value '--watch=false' will be used.
  Error: Schema validation failed with the following errors:
    Data path "" must NOT have additional properties().
```
I suppose streamlining the number or the uniqueness of the environments would be good.

lukestanley:
That's all it took to fix it. This looks good.

lukestanley (approved):

PR #350, opened 2023-01-12
Service worker bypass and http block doc

the-greenman:
This is a tiny update to implement the service worker bypass flag, and make a few improvements to the http block doc. 

Rather than making another config option, I've implemented the header as standard. I don't think it makes too much sense to add yet another option. In our case, I think it's a going to be handy to bake this in.


lukestanley (approved):
:+1: :tada: nice sleuthing.

Issue #364, opened 2023-06-07
Document custom Jmespath mappings


lukestanley:
Including new UNIX epoch timestamp converter, Excel timestamp converter and # other custom extensions to the Jmespath syntax.

PR #365, opened 2023-06-08
Documents JMESPath extensions

lukestanley:
This documents each custom JMESPath extension we have with examples.
I already tested the documentation renders correctly on readthedocs by commiting to master to test it is rendering correctly and we should now sanity check and properly review it as part of #364
This is how it rendered: https://kendraio-app.readthedocs.io/en/latest/workflow/blocks/mapping.html


CodeKrakken (approved):

lukestanley:
This was for retrospective approval, I now remember! So there is nothing more to be done now it's approved.

PR #389, opened 2023-09-21
Adds a script to find examples of particular blocks using specific properties from Flow Cloud

lukestanley:
We often need to look at examples of Flows using specific blocks, and sometimes specific config properties on them - like Forms using "uiSchema". I (well, and GPT-4) made this to help make that process much faster.

### DIFF
--- a/docs/extract-example-block-usage-from-flows.js
+++ b/docs/extract-example-block-usage-from-flows.js
@@ -13,8 +13,8 @@
  */
 
 
-let inputFilePath = 'flows.json';
-let outputFilePath = 'flow-block-config-analysis.json';
+let inputFilePath = 'flow-analysis/flows.json';
+let outputFilePath = 'flow-analysis/flow-block-config-analysis.json';
 let blockType = 'form';  // Replace with the block type you're interested in
 let propertyOfInterest = null;  // E.g: 'uiSchema' - replace with the property you're interested in or leave as null
 
--- a/docs/extract-example-block-usage-from-flows.js
+++ b/docs/extract-example-block-usage-from-flows.js
@@ -38,9 +38,22 @@ const generateGroupedJson = (inputFilePath, outputFilePath, blockType, propertyO
   for (const flow of flowsData) {
     let flowBlocks = [];  // To store the blocks with the property of interest for this flow
 
+
     if (flow.blocks) {
       for (const block of flow.blocks) {
-        if (block.type === blockType && (!propertyOfInterest || (block[propertyOfInterest] && Object.keys(block[propertyOfInterest]).length > 0))) {
+        const hasMatchingBlock = block.type === blockType;
+        let hasPropertyOfInterest;
+        // If no property of interest is specified, so all properties are interesting!
+        if (!propertyOfInterest) {
+          hasPropertyOfInterest = true;
+        }
+
+        if (propertyOfInterest) {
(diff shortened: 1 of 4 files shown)


docs/extract-example-block-usage-from-flows.js
```diff
@@ -0,0 +1,93 @@
...
+  let enhancedFlowsWithBlocks = [];
+  let groupedByAdapter = {};
+
+  // Iterate through the flows to collect blocks and their metadata
+  for (const flow of flowsData) {
+    let flowBlocks = [];  // To store the blocks with the property of interest for this flow
+
+    if (flow.blocks) {
+      for (const block of flow.blocks) {
+        if (block.type === blockType && (!propertyOfInterest || (block[propertyOfInterest] && Object.keys(block[propertyOfInterest]).length > 0))) {
```
gsambrotta, line 42: This is quite hard to read, at least for me. 
Can we at least save some of the logic in a var so is a bit more readable? 
Like `const emptyPropertOfInterest = block[propertyOfInterest] && Object.keys(block[propertyOfInterest]).length > 0`
and `const propertyOfInterestIsFalse = !propertyOfInterest || emptyPropertOfInterest`

what do you think @lukestanley ?

docs/extract-example-block-usage-from-flows.js
```diff
@@ -0,0 +1,93 @@
...
+      }
+    }
+
+    if (flowBlocks.length > 0) {
+      // Add metadata
+      const blockMetadata = {
+        adapterName: flow.adapterName,
+        id: flow.id,
+        title: flow.title,
+        url: `https://app.kendra.io/${flow.adapterName}/${flow.id}`
```
gsambrotta, line 54: Can we save static Url in a var on top of the file?
Like `const KENDRAIO_APP_URL = https://app.kendra.io/`

docs/extract-example-block-usage-from-flows.js
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * How to Use:
+ * 1. Modify the variables below to specify the block type and property of interest (if any).
+ * 2. Run the script using Node.js.
+ * 3. Check the generated JSON file for the grouped data.
+ */
+
+
+let inputFilePath = 'flows.json';
+let outputFilePath = 'flow-block-config-analysis.json';
```
gsambrotta, line 17: I don't find it clear where the files will be saved/outputted.
Also I don't think `root` is the best place. What about create a `flow-analysis` folder and keep these files there? 
And then we can hide from git so we don't deploy it.

gsambrotta (changes requested):
Niiiiceee! :star2:  
This is exactly what we need! Thank you very much for working on this 
Code generally looks good to me, just some small comments about code style.

docs/extract-example-block-usage-from-flows.js
lukestanley: Done @gsambrotta

gsambrotta:
Looks good to me now :clap: :clap: 

Don't we want to add this script to the docs? We could add a section like "utility" or "helpful tools" and add this there.
What do you think @lukestanley ? I think would be a pity if this scripts stays hidden in the code (and probably get forgotten soon)

lukestanley:
Ready for review @gsambrotta, thanks.

gsambrotta (approved):
Awesome!! :clap: :clap:

PR #392, opened 2023-10-04
Return error information on HTTP block request

gsambrotta:
Currently, the HTTP block handle error in the code but it does not return any information about it.
If a GET/POST/PUT/... request is made and an error is raised, the data object in the flow is `null`

With this PR the data object in the flow, will now return the error information.
It also contain a very handy `hasError: true` property so it can be easily check for error or not and potentially display error messages in the flow.

Here two screenshots of how the result will be:
image
image

And here is a flow that can be use to test this PR: https://app.kendra.io/harvest-timesheet/harvestTimesheet

### DIFF
--- a/docs/workflow/blocks/form.rst
+++ b/docs/workflow/blocks/form.rst
@@ -44,7 +44,7 @@ not dependent on the form, then it is best to place them before the form in the 
 
 Config init
 -----------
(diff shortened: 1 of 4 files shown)


gsambrotta:
Closing this has it has the wrong git history in it

PR #393, opened 2023-10-04
Return error information on HTTP block request

gsambrotta:
Currently, the HTTP block handle error in the code but it does not return any information about it.
If a GET/POST/PUT/... request is made and an error is raised, the data object in the flow is `null`

With this PR the data object in the flow, will now return the error information.
It also contain a very handy `hasError: true` property so it can be easily check for error or not and potentially display error messages in the flow.

Here two screenshots of how the result will be:
image
image

And here is a flow that can be use to test this PR: [Kendraio Harvest Time](https://app.kendra.io/harvest-timesheet/harvestTimesheet)

Linear Issue [K-361](https://linear.app/kendraio/issue/K-361/kendraio-time-flow)

### DIFF
--- a/cypress/e2e/http_block.ts
+++ b/cypress/e2e/http_block.ts
@@ -81,13 +81,14 @@ describe('HTTP Block Request', () => {
             },
             {
                 "type": "debug",
-                "open": 2,
+                "open": 3,
                 "showData": true
             }
         ]);
 
-        console.log
-        cy.contains('hasError');
+        cy.contains('hasError:true');
+        cy.contains('status:400');
+        cy.contains('errorMessage:"Http failure response for https://example.com/data: 400 Bad Request"');
         cy.get('app-template-block').contains('Error with submission')
     });
 
--- a/cypress/e2e/http_block.ts
+++ b/cypress/e2e/http_block.ts
@@ -3,7 +3,100 @@ import { loadFlowCode } from '../support/helper';
 // tslint:disable: quotemark
 /// <reference types="Cypress" />
 
-describe('HTTP Block Integration Tests', () => {
+
+describe('HTTP Block Request', () => {
+
+    beforeEach(() => {
+        // Prevent external network request for adapter config
+        cy.intercept('GET', 'https://kendraio.github.io/kendraio-adapter/config.json', {
+            fixture: 'adapterConfig.json'
+        });
+
+        // Prevent external network requests for Workflow cloud
+        cy.intercept('GET', 'https://app.kendra.io/api/workflowCloud/listWorkflows', {
+            fixture: 'workflow-cloud.json'
+        });
+
+        // Prevent external network requests for fonts with empty CSS rule 
+        cy.intercept('https://fonts.googleapis.com/\*\*', "\*{ }");
+    });
+
+    it('should return a single set of results. Without pagination', () => {
+        cy.intercept({
+            url: 'https://example.com/data'
+        }, {
+            statusCode: 200,
+            body: '["hippo", "giraffe"]'
+        });
+
+        loadFlowCode([
+            { "type": "init" },
+            {
+                "type": "http",
+                "method": "GET",
+                "endpoint": "https://example.com/data"
+            },
+            {
+                "type": "debug",
+                "open": 2,
+                "showData": true
+            }
+        ]);
+        cy.contains('hippo');
+        cy.contains('giraffe');
+    });
+
+    it('should return an error', () => {
+        cy.intercept({
+            url: 'https://example.com/data'
+        }, {
+            statusCode: 400,
+            body: { error: {
+                        error: "Http failure 400 Bad request",
+                        error_description: "There was a problem with your request" 
+                    }   
+                }
+        });
+
+        loadFlowCode([
+            { "type": "init" },
+            {
+                "type": "http",
+                "method": "GET",
+                "endpoint": "https://example.com/data",
+                "onError": {
+                    "blocks": [
+                        {
+                            "type": "card",
+                            "blocks": [
+                                {
+                                    "type": "template",
+                                    "template": "Error with submission:<p>{{data.error.error}} - {{data.error.error_description}}</p>"
+                                }
+                            ]
+                        }
+                    ]
+                }
+
+            },
+            {
+                "type": "debug",
+                "open": 2,
+                "showData": true
+            }
+        ]);
+
+        console.log
+        cy.contains('hasError');
+        cy.get('app-template-block').contains('Error with submission')
+    });
+
+
+});
+
+
+
+describe('HTTP Block Follow Pagination', () => {
 
     beforeEach(() => {
         // Prevent external network request for adapter config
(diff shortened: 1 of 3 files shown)


lukestanley:
This seems good, nice work! It needs tests updating though:
see: `cypress/e2e/http_block_page_following.ts`
@gsambrotta

cypress/e2e/http_block.ts
```diff
@@ -3,7 +3,100 @@ import { loadFlowCode } from '../support/helper';
 // tslint:disable: quotemark
 /// <reference types="Cypress" />
 
-describe('HTTP Block Integration Tests', () => {
+
```
gsambrotta, line 6: Unfortunately changing the file name made code review a bit more difficult.
To help here some info: 
- I divided the test in two test suits: a normal http request and paginated request
- The new code is inside the first test suit: `HTTP Block Request`

lukestanley (changes requested):
We need to see that the error code and body can be used too. I don't think this does that, right?
@gsambrotta

gsambrotta:
> We need to see that the error code and body can be used too. I don't think this does that, right? @gsambrotta

@lukestanley It does in the `onError` config. https://github.com/kendraio/kendraio-app/pull/393/files#diff-5da5ddc4e5c0ad1ea378ff3988f2537b9975bfc9c90fb9ba8d31a75ee323858eR74
Here I use `data.error.error_message` to display the message. Obviously it can be use also `data.error.statusCode`

For some reasons it does not get display in the Cypress view, so I cannot `cy.get` like I do with the error static message.
I think the data get populated after the card is created and it doesn't display the message.

Do you think is not clear like it is now? Is needed to check the existence of the error message or status code?

docs/workflow/blocks/http.rst
```diff
@@ -50,7 +51,8 @@ For simple requests, the ``endpoint`` can just be a simple string:
     }
 
 
-If the endpoint needs to be constructed from data, the endpoint can be specified as an object with a "valueGetter" attribute. 
+**Dynamic data** If the endpoint needs to be constructed from data, the endpoint can be specified as an object with a "valueGetter" attribute.
+"valueGetter" get data just from context.
```
CodeKrakken, line 55: Better English would be - 

"valueGetter" can only get data from the context.

CodeKrakken (changes requested):
Just that one syntax comment. Code looks clear and all tests are passing.

docs/workflow/blocks/http.rst
gsambrotta: Thanks!

docs/workflow/blocks/http.rst
```diff
@@ -62,15 +64,32 @@ If the endpoint needs to be constructed from data, the endpoint can be specified
...
-Custom headers can also be specified using JMES Path expressions:
+Custom headers can also be specified using JMESPath expressions.
+Be careful to write your header's value with two types of quote (double quotes + single quotes), if the value is a string.
 
 .. code-block:: json
 
   {
       "type": "http",
-      "method": "post",
+      "Defaultmethod": "post",
```
lukestanley, line 92: What's going on here? @gsambrotta

gsambrotta: Ah, thanks! I think I was typing somewhere else, and it accidentally typed on that line.

docs/workflow/blocks/http.rst
```diff
@@ -33,12 +33,13 @@ Supported properties
   property.
 - **headers** - A set of headers with header name as object key. Values are processed by JMESpath
 - **endpoint** - The request endpoint. Can take multiple forms. See below. 
+- **onError** - It expects blocks as argument. These blocks will be visible in case of HTTP request error.
```
lukestanley, line 36:
```suggestion
- **onError** - Define an array of blocks to show when there is an error processing the HTTP request. 
```

docs/workflow/blocks/http.rst
lukestanley, line 92:
```suggestion
      "method": "post",
```

docs/workflow/blocks/http.rst
```diff
@@ -62,15 +64,32 @@ If the endpoint needs to be constructed from data, the endpoint can be specified
...
+        }
+    }
+  }
+
 
+**Headers** 
 For advanced use cases, the payload can be constructed using a JMES Path expression.
-Custom headers can also be specified using JMES Path expressions:
+Custom headers can also be specified using JMESPath expressions.
+Be careful to write your header's value with two types of quote (double quotes + single quotes), if the value is a string.
```
lukestanley, line 86: This part needs an example probably.
I tried to improve it as is though:
```suggestion
Caution: if the custom header value is a string, header values must use two types of quotes: double quotes and single quotes.
```

gsambrotta: Yeah, you are right, is not clear, but I don't know how to explain it. 
I am not sure about your suggestion, neither, though.
Why did you write `custom header`? This tip (double+single quotes) is valid for any headers, custom or not. Or when we say "custom header" we mean "non-dynamic value"?

There is an example, is right after this text, in the http block example. https://github.com/kendraio/kendraio-app/pull/393/files#diff-544912a428316debac6dc397877aae4d88c8bdac4836974e3caa2c8389ffc96dR98

But yeah, maybe we can be more explicit and add this line `"payload": "'grant_type=client_credentials'"` in the explanation.

Something like:
> Caution: if the header value is a string, the header value must use two types of quotes: double quotes and single quotes like in the following example: `"payload": "'grant_type=client_credentials'"`

What do you think @lukestanley ?

Actually, I have no idea if it should be in double + single quotes in the case the value is a number or boolean. 
But I don't think we will ever have a header that is not either a string or a dynamic value

lukestanley: Actually I'd replace this:
`For advanced use cases, the payload can be constructed using a JMES Path expression.
Custom headers can also be specified using JMES Path expressions:
Custom headers can also be specified using JMESPath expressions.`
With:

`JMESPath expressions can be used to dynamically set header values.` Rather than the duplication. This would be more clear.

And thanks for pointing out that there is an example below. @gsambrotta

gsambrotta: Good catch to avoid repetition. I think it still good to mention the payload tho. 
What about this text for the whole description:
```
JMESPath expressions can be used to dynamically set header and payload values.
Caution: if the header value is a string, it must use two types of quotes: double and single quotes, like "payload": "'grant_type=client_credentials'".
```
Is this good for you @lukestanley ?

lukestanley: I thumbed up suggested text last week. So a commit with that in it is needed, right?  @gsambrotta

gsambrotta:
@lukestanley I committed the text as we agreed. Would you please give another check?

CodeKrakken (commented):
@gsambrotta That all looks good to me.

gsambrotta:
> @gsambrotta That all looks good to me.

@CodeKrakken If so, can you please approve it?

CodeKrakken:
> > @gsambrotta That all looks good to me.
> 
> @CodeKrakken If so, can you please approve it?

Sorry - was waiting for @lukestanley to review it.

gsambrotta:
@CodeKrakken ahh you are right. I thought we configure to have at least 1 approval but apparently an approval from anyone that has done the review is needed.

lukestanley (approved):
Looks good! Well done! @gsambrotta

PR #394, opened 2023-10-04
Update documentation - Form Block

gsambrotta:
Some small but helpful improvements in Kendraio Docs about Form block
- [x] read-only
- [x] example how a value can be used as title
- [x] example value with backticks: set string as value in form/mapping i.e. "type": "string" otherwise it return null i.e - Added to the Mapping doc as it is more general to JMESPath
- [ ] validation

Linear ticket [K-332](https://linear.app/kendraio/issue/K-332/update-form-documentation)

### DIFF
--- a/docs/workflow/blocks/form.rst
+++ b/docs/workflow/blocks/form.rst
@@ -44,16 +44,74 @@ not dependent on the form, then it is best to place them before the form in the 
 
 Config init
 -----------
-It can often be beneficial to initialiase the form data by creating a mapping of the default data. 
+The initial configuration will result in an empty form.
+The fields of the form can be populated in two ways: manually writing fields or dynamically generating fields from given data.
 
-.. code-block:: json 
+Here's an example of creating a form by manually writing fields and configuring it:
 
+.. code-block:: json
   {
-    "confirm": false
+    "type": "form",
+    "hasSubmit": "true", 
+    "label": "Click to submit",
+    "jsonSchema": {
+        "type": "object",
+        "properties": {
+            "name": {
+                "type": "string",
+                "title": "Name",
+                "default": ""
+            },
+            "lastname": {
+                "type": "string",
+                "title": "Last name",
+                "default": ""
+            }
+        }
+    },
+    "uiSchema": {}
+}
+
+Dynamic data and field titles
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+To generate fields form given data, the data needs to be in a format readable by the form block.
+It is possible to transform the data into the format that the form expects with the help of a mapping block.
+Is also possible to use generated data to dynamically display the title of a field
+
+.. code-block:: json
+  
+
+  // generated data
+  {
+    "name": `John`,
+    "surname": `Doe`
   }
 
-Examples
----------
+  // Form config
+  {
+    "type": "form",
+    "jsonSchema": {
+        "type": "object",
+        "properties": {
+            "title": {
+                "type": "string",
+                "title": "I am harcoded text"
+            },
+            "user_name": {
+                "type": "string",
+                "title": "name",
+                "default": "The title of this field comes from dynamic data"
+            },
+            "user_surname": {
+                "type": "string",
+                "title": "surname",
+                "default": "The title of this field comes from dynamic data"
+            }
+        }
+    },
+    "uiSchema": {}
+}
+
 
 Read-only
 ^^^^^^^^^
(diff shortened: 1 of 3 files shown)


gsambrotta:
@lukestanley  @the-greenman  Do you have any idea/information about the validation of single fields? I think I have seen it but I can't recall nor find again in which flow.

docs/workflow/blocks/form.rst
```diff
@@ -44,21 +44,83 @@ not dependent on the form, then it is best to place them before the form in the
 
 Config init
 -----------
-If can often be beneficial to initialiase the form data by creating an mapping of the default data. 
+It can often be beneficial to initialiase the form data by creating a mapping of the default data.
```
lukestanley, line 47: I noticed 'initialiase' is not spelt correctly and that it wasn't clear English then I came up with this:
```suggestion
To fill out existing information easily, a form can be loaded with pre-existing fields, using a 'mapping' - a rule that determines how to generate the default data.
```

docs/workflow/blocks/form.rst
```diff
@@ -44,21 +44,83 @@ not dependent on the form, then it is best to place them before the form in the
...
+    },
+    "uiSchema": {}
+  }
+
+Form validation
+^^^^^^^^^^^^^^^
+
+Dynamic field title
+^^^^^^^^^^^^^^^^^^^
+Using dynamic data as title for the form field
```
lukestanley, line 85:
```suggestion
Using dynamic input data, as title for the form field
```

docs/workflow/blocks/form.rst
```diff
@@ -44,21 +44,83 @@ not dependent on the form, then it is best to place them before the form in the
...
+        "type": "object",
+        "properties": {
+            "title": {
+                "type": "string",
+                "title": "I am harcoded text"
+            },
+            "user_name": {
+                "type": "string",
+                "title": "name",
+                "default": "titlte of this field come from dynamic data"
```
lukestanley, line 109:
```suggestion
                "default": "The title of this field comes from dynamic data"
```

docs/workflow/blocks/form.rst
```diff
@@ -44,21 +44,83 @@ not dependent on the form, then it is best to place them before the form in the
...
+            },
+            "user_surname": {
+                "type": "string",
+                "title": "surname",
+                "default": "titlte of this field come from dynamic data"
```
lukestanley, line 114:
```suggestion
                "default": "The title of this field comes from dynamic data"
```

docs/workflow/blocks/mapping.rst
```diff
@@ -53,9 +53,39 @@ read from *filter*. Reading from  *state.global.workflowCloud.listWorkflows.filt
 Handy JMESPath patterns
 -----------------------
 
+Using string as value
```
lukestanley, line 56:
```suggestion
Using a string as a value
```

docs/workflow/blocks/mapping.rst
```diff
@@ -53,9 +53,39 @@ read from *filter*. Reading from  *state.global.workflowCloud.listWorkflows.filt
...
+^^^^^^^^^^^^^^^^^^^^
+When manually creating data, a string will be passed as value.
```
lukestanley, line 58:
```suggestion
Data can be made manually using a string value.
```

docs/workflow/blocks/mapping.rst
```diff
@@ -53,9 +53,39 @@ read from *filter*. Reading from  *state.global.workflowCloud.listWorkflows.filt
...
+Usually keys and values are wrapped in double quotes, to be valid JMESPath.
```
lukestanley, line 59: This is unnecessarily vague. Usually, or always? If not always, then when? @gsambrotta

docs/workflow/blocks/mapping.rst
```diff
@@ -53,9 +53,39 @@ read from *filter*. Reading from  *state.global.workflowCloud.listWorkflows.filt
...
+
+Output will be:
+
+.. code-block:: text
+
+  name:"John"
+  surname:"Doe"
+  iAmNull:null
+
+Is also possible to wrap the whole expression in backtick. The output will be the same:
```
lukestanley, line 77:
```suggestion
We've seen that property values of object keys can be set using backticks, and it is also possible to wrap a whole JSON object in backticks. The output will be the same as this:
```

docs/workflow/blocks/form.rst
gsambrotta: Mhm honestly both are not clear to me. If I would be new to Kendraio project I would have no idea what we are talking about :D 
I suggest to add an example that is more helpful then 
```

  {
    "confirm": false
  }

```
I will commit the changes and then we can discuss on the new text/example

docs/workflow/blocks/mapping.rst
gsambrotta: mhm I see your point. I rewrite the whole bit. Maybe now is more clear.

gsambrotta:
@lukestanley FYI I have correct the text following your suggestion and I have rewrite some parts.

docs/workflow/blocks/mapping.rst
gsambrotta: I rewrote it in a different way, let me know if is understable

lukestanley (approved):

PR #398, opened 2023-10-23
Npm security check

gsambrotta:


CodeKrakken (commented):
I have no issues with the code, as long as it works. Probably best @lukestanley takes a look too though, as my Cypress experience is limited.

cypress.config.ts
```diff
@@ -1,30 +1,35 @@
...
-  screenshotsFolder: 'cypress/screenshots',
-  fixturesFolder: 'cypress/fixtures',
-  reporter: 'mochawesome',
+  screenshotsFolder: "cypress/screenshots",
+  fixturesFolder: "cypress/fixtures",
+  reporter: "mochawesome",
+
   reporterOptions: {
-    reportDir: 'cypress/results',
+    reportDir: "cypress/results",
```
lukestanley, line 11: Please can this sorta linting be a separate commit so we can more easily tell what is going on, other than linting? @gsambrotta

cypress/support/component.ts
```diff
@@ -0,0 +1,39 @@
...
+// This example support/component.ts is processed and
+// loaded automatically before your test files.
+//
+// This is a great place to put global configuration and
+// behavior that modifies Cypress.
+//
+// You can change the location of this file or turn off
+// automatically serving support files with the
+// 'supportFile' configuration option.
+//
```
lukestanley, line 11: Do we need this entire auto-added component example? I guess I'm okay with it, but I wanted to check you know that it's here, just in case it was added accidentally. @gsambrotta

lukestanley (approved):
In the interest of making progress I'm approving this even though I'm not sure about the extra example Cypress files, I can live with them.

lukestanley:
Nice work on this BTW. @gsambrotta

cypress.config.ts
gsambrotta: Oh yes, sorry, I didn't see it. It was automatically corrected by Cypress CLI.

cypress/support/component.ts
gsambrotta: It was autogenerated by Cypress when I updated with the CLI. 
I can remove it. Let's keep it clean :)

gsambrotta: @lukestanley I decided to delete all the support folder. This folder was created automatically by the CLI when I added the option to configure for testing Angular components. But as we do not have any tests for components, the tests were broken. 
I therefore deleted the configuration and thought it was good to have an example there anyway.
One a second thought, I don't think it makes sense. If we decide to create a test for component, we can easily regenerate this folder

gsambrotta:
@lukestanley I decided to delete all the `support` folder. This folder was created automatically by the CLI when I added the option to configure for testing Angular components. But as we do not have any tests for components, the tests were broken. 
I therefore deleted the configuration and thought it was good to have an example there anyway.
One a second thought, I don't think it makes sense. If we decide to create a test for component, we can easily regenerate this folder

lukestanley:
Cool, I agree with that assessment. Nice work. @gsambrotta

Issue #399, opened 2023-10-25
Improve flexbox style in Batch block


gsambrotta:
Currently, we can use Batch block as a wrapper to display several items.
Is possible to display those items with flexbox style, but currently the items will always be displayed with 
`justify-content: space-between` which is not always what we want.

In particular, if the items are not broad, the design will look off.

gsambrotta:
Close as PR was merged #400

PR #400, opened 2023-10-30
Improve flexbox batch block

gsambrotta:
PR for the Issue #399

### DIFF
--- a/src/app/blocks/batch-block/batch-block.component.scss
+++ b/src/app/blocks/batch-block/batch-block.component.scss
@@ -1,5 +1,4 @@
-.batch-block-flex {
-    background-color: black;
+:host(.batch-block-flex) {
     display: flex;
     flex-wrap: wrap;
     justify-content: flex-start;
--- a/src/app/blocks/batch-block/batch-block.component.scss
+++ b/src/app/blocks/batch-block/batch-block.component.scss
@@ -1,0 +1,20 @@
+.batch-block-flex {
+    background-color: black;
+    display: flex;
+    flex-wrap: wrap;
+    justify-content: flex-start;
+    align-items: center;
+    gap: 10px 20px;
+    overflow: hidden;
+}
+
+.multi-wrapper {
+  flex: 1;
+  overflow: hidden;
+  min-width: 250px;
+  max-width: 50%;
+
+  &:last-child {
+    flex-shrink: 0;
+  }
+}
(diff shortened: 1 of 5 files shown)


lukestanley (changes requested):
This looks good.
Why hide overflow? Is that better than allowing scrolling?
I think it would be good to add to the documentation more about the block and maybe mention other blocks it's good with?
A bit more detail about it's current usage that's evolved might help.

gsambrotta:
> Why hide overflow? Is that better than allowing scrolling?
I hide the overflowing content of the whole container, so not specific cards. It doesn't actually overflow, but is a precaution to be sure that all the cards will fit in the container.
In this situation, I will avoid scrolling. The cards have to resize or fall in another row. 
I would say we never want the situation in which a user has to scroll to see more cards. Why? It is ugly and not user-friendly at all.

lukestanley:
If the `overflow: hidden` property it's just a precaution, for a situation that we've not ran into, isn't it best to not have it there and keep the CSS more minimal? @gsambrotta

src/app/blocks/batch-block/batch-block.component.scss
```diff
@@ -0,0 +1,19 @@
+:host(.batch-block-flex) {
+    display: flex;
+    flex-wrap: wrap;
+    justify-content: flex-start;
+    align-items: center;
+    gap: 10px 20px;
+    overflow: hidden;
```
lukestanley, line 7: If the `overflow: hidden` property it's just a precaution, for a situation that we've not ran into, isn't it best to not have it there and keep the CSS more minimal? @gsambrotta

src/app/blocks/batch-block/batch-block.component.scss
```diff
@@ -0,0 +1,19 @@
...
+}
+
+.multi-wrapper {
+  flex: 1;
+  overflow: hidden;
```
lukestanley, line 12: Same thing here, is this needed? @gsambrotta

src/app/blocks/batch-block/batch-block.component.scss
gsambrotta: Yeah, I was also battling with myself to decide whether I should leave it or not. 
Let's remove it. It should not make any difference.

lukestanley (approved):
Lovely stuff.

PR #407, opened 2023-11-06
Merge HTTP block error feedback, Batch block flexbox improvment for cards, block research tool with documentation, upgrades (PR's #393, #400, #389, #398)

lukestanley:
I think the title covers it and it should be good to put into a commit and description.


gsambrotta (approved):
:+1: :+1:

PR #408, opened 2023-11-09
Fix card's style

gsambrotta:
- remove padding around image
- make image take all the container's space
- move card style in card style's component
- add truncate helper class and documentation in style.scss

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,8 +1,13 @@
 /*********
 Global style guide
-_ If a block is nested or the HTML is generated, the style will not be applicable to the component style and will need to be written here
-_ There are some Helpers class among this file. Search for "HELPER CLASS" to find them.
-_ The style attribute does not work on the HTML block, as the HTML is sanitize. If is needed to style a specific HTML tag, is possible to add a class and add the style here.
+/*********
+Global Style Guide:
+- Styles not applicable directly within component styles should go here.
+- Helper classes are listed below. They are usable within sanitised
+  HTML and are marked as "HELPER CLASS" for easy identification.
+- Helper classes are particularly useful with the Template block, 
+  where the style attribute is sanitised, but CSS classes are allowed.
+*********/
 *********/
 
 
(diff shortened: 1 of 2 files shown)


src/styles/main.scss
```diff
@@ -1,3 +1,11 @@
+/*********
+Global style guide
+_ If a block is nested or the HTML is generated, the style will not be applicable to the component style and will need to be written here
+_ There are some Helpers class among this file. Search for "HELPER CLASS" to find them.
+_ The style attribute does not work on the HTML block, as the HTML is sanitize. If is needed to style a specific HTML tag, is possible to add a class and add the style here.
+*********/
```
CodeKrakken, line 6: A few grammar/spelling bits here. Here's a corrected version:

Global style guide
_ If a block is nested or the HTML is generated, the style will not be applicable to the component style and will need to be written here.
_ There are some helper classes below. Search for "HELPER CLASS" to find them.
_ The style attribute does not work on the HTML block, as the HTML is sanitised. If you need to style a specific HTML tag, you can add a class for the style here.

CodeKrakken (changes requested):
Just one comment on the English text. Otherwise no problems - looks very nice!

src/styles/main.scss
```diff
@@ -1,3 +1,11 @@
+/*********
+Global style guide
+_ If a block is nested or the HTML is generated, the style will not be applicable to the component style and will need to be written here
+_ There are some Helpers class among this file. Search for "HELPER CLASS" to find them.
+_ The style attribute does not work on the HTML block, as the HTML is sanitize. If is needed to style a specific HTML tag, is possible to add a class and add the style here.
```
lukestanley, line 5:
```suggestion
/*********
Global Style Guide:
- Styles not applicable directly within component styles should go here.
- Helper classes are listed below. They are usable within sanitised
  HTML and are marked as "HELPER CLASS" for easy identification.
- Helper classes are particularly useful with the Template block, 
  where the style attribute is sanitised, but CSS classes are allowed.
*********/
```
This is more clear and concise. @gsambrotta

lukestanley (approved):
Approved with a suggested change for the big comment.

CodeKrakken (approved):

PR #414, opened 2023-11-24
Adds YouTube triplet privacy setting

lukestanley:
YouTube require selection from 3 different privacy settings.

### DIFF
--- a/src/app/components/youtube-upload/youtube-upload.component.html
+++ b/src/app/components/youtube-upload/youtube-upload.component.html
@@ -26,7 +26,11 @@
 
         <input type="file" placeholder="Select file" (change)="onFileChange($event)">
 
-        <mat-checkbox formControlName="isPrivate">Private</mat-checkbox>
+        <mat-radio-group formControlName="privacySetting">
+          <mat-radio-button value="private">Private</mat-radio-button>
+          <mat-radio-button value="public">Public</mat-radio-button>
+          <mat-radio-button value="unlisted">Unlisted</mat-radio-button>
+        </mat-radio-group>
       </form>
     </ng-container>
 
--- a/src/app/components/youtube-upload/youtube-upload.component.ts
+++ b/src/app/components/youtube-upload/youtube-upload.component.ts
@@ -40,7 +40,7 @@ export class YoutubeUploadComponent implements OnInit {
       description: ['', [Validators.required]],
       category: ['', [Validators.required]],
       file: [null, [Validators.required]],
-      isPrivate: [true]
+      privacySetting: ['unlisted', [Validators.required]]
     });
     this.categories$ = this.yt.getCategories();
     this.progress$ = this.yt.progress$;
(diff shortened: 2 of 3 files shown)


lukestanley:
Yeah I think it's what they require, and from localhost it works. I'll probably investigate the wildcard hostname situation Monday.

lukestanley:
To test, I suggest making some random test videos:
```
# This should work on an Ubuntu system:
# sudo apt install python3 python3-pip ffmpeg -y && pip3 install numpy pillow
# Run this in with python3:
import numpy as np
from PIL import Image
import os

def generate_random_video(video_number, width=1280, height=720, duration=10):
    # Generate a random image
    array = np.random.randint(256, size=(height, width, 3), dtype=np.uint8)
    image = Image.fromarray(array)

    # Save the image to your local directory
    image_path = f"/tmp/random_image_{video_number}.png"
    image.save(image_path)

    # Define the video path
    video_path = f"/tmp/{video_number}.mp4"

    # Convert the image to a video using ffmpeg
    ffmpeg_command = f"ffmpeg -loop 1 -i {image_path} -c:v libx264 -t {duration} -pix_fmt yuv420p -vf scale=1280:720 {video_path}"
    os.system(ffmpeg_command)

for i in range(12):
    generate_random_video(i)

```

Login to a Google account that has YouTube here: 
https://kendraio-app-git-youtube-privacy-setting-kendraio.vercel.app/user
Do the auth flow as needed.

When done with the login, upload the test videos here:
https://kendraio-app-git-youtube-privacy-setting-kendraio.vercel.app/youtube

Test each state: private, public and unlisted.

gsambrotta:
I am trying to review this, but I don't understand how to test it. Actually, I do not even understand the goal of this flow. I guess that is to list YouTube videos. And I do not know which inputs I need (what is a snippet? A YouTube URL?) in order to test this flow. 
I think we can do better in writing descriptions and how to test instructions in our PRs, so the other people don't have to spend time figuring out what it is and how to do it. 
I read your comment @lukestanley about how to test it, but honestly, it confuses me even more. Why do I need to create random videos? And why with Python? Once again, I guess, I miss some context.


The code looks good. If @dahacouk approves the visual, I am happy to approve the code.

lukestanley:
It's for uploading videos, not just listing them! So we need some videos to upload.
I've seen YouTube block uploads of videos it already has, hence the random videos! It's a quick helper script, you should just be able to run it. It should run easily. @gsambrotta

lukestanley:
I'm not sure where you saw "snippet"? But it usually means a small bit of code or text. Possible a sample of a larger bit of code or text. For testing YouTube uploads, it needs a title, a number works. It also might appear hung during uploading but the test videos are tiny, and should actually upload really fast. I looked at https://studio.youtube.com to check the privacy status of each video I uploaded. @gsambrotta

gsambrotta (approved):
:tophat:

PR #421, opened 2023-12-21
Adds toUppercase and replaceAll functions

CodeKrakken:
This update extends JMESPath functionality to include replaceAll and toUpper functions, along with tests.

The docs have also been updated with usage instructions for the new functions. There is some minor rephrasing of related functions in the interest of linguistic consistency.

### DIFF
--- a/docs/workflow/blocks/mapping.rst
+++ b/docs/workflow/blocks/mapping.rst
@@ -273,31 +273,33 @@ Available Extensions
 3. findIndex
 4. uuid
 5. toLower
-6. replace
-7. trim
-8. now
-9. formatDate
-10. omit
-11. pick
-12. split
-13. find
-14. compact
-15. qs
-16. parseQs
-17. zip
-18. debug
-19. json
-20. markdown
-21. btoa
-22. base64encode
-23. pairwise
-24. numDiff
-25. percentChange
-26. groupByKeys
-27. all
-28. parseDate
-29. parseDuration
-30. parseUnixTimestamp
+6. toUpper
+7. replace
+8. replaceAll
+9. trim
+10. now
+11. formatDate
+12. omit
+13. pick
+14. split
+15. find
+16. compact
+17. qs
+18. parseQs
+19. zip
+20. debug
+21. json
+22. markdown
+23. btoa
+24. base64encode
+25. pairwise
+26. numDiff
+27. percentChange
+28. groupByKeys
+29. all
+30. parseDate
+31. parseDuration
+32. parseUnixTimestamp
 
 Each of these extensions is detailed below with example usages.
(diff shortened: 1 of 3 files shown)


lukestanley (changes requested):
It would be good to have a description that describes all the changes you intend for the PR to introduce. There is more going on here than what the title suggests. It might seem obvious but it would be good for the tests to point out how the existing replace function works, for example (just a sketch): `"rulers:cat,cat".replace('cat','dog')=="rulers:dog,cat"`
By the way, I had to hide whitespace changes in the PR diff viewer, since there were indentation changes. Auto-formatted changes can really made understanding changes harder than need be. I'm glad this was easy to do in this PR view, but in the future, could we agree to keep indentation changes to a separate piece of work?
@CodeKrakken

CodeKrakken:
No problem @lukestanley. I've added to the description and I agree about indentation - it's distracting amongst other changes. As for pointing out how the functions work, how about the below example?

`it('should replace one string instance within a larger string', () => {

    const data = { string: "test string before" };
    const expr1 = "replace(string, 'before', 'after')";
    const expected1 = "test string after";

    // "test string before".replace('before','after')=="test string after"

    expect(mappingUtility(data, expr1)).toBe(expected1);
  });`

lukestanley:
Did you expect to commit a test for trim too? I didn't see it in the description.

> 
> ```
> it('should replace one string instance within a larger string', () => {
> 
>     const data = { string: "test string before" };
>     const expr1 = "replace(string, 'before', 'after')";
>     const expected1 = "test string after";
> 
>     // "test string before".replace('before','after')=="test string after"
> 
>     expect(mappingUtility(data, expr1)).toBe(expected1);
> });
> ```

No, I meant a test to make it obvious how .replace only replaces one item.
@CodeKrakken

CodeKrakken:
I've removed the trim test - it's not relevant to this code change. I'll open an issue for increasing test coverage so we don't lose track of it.

I've updated my replace and replaceAll tests - they should be clearer now.

The numbers - do you mean `expr1` and `expected1`? I was just matching the existing style. I also don't see the need for them - shall we remove them? @lukestanley

lukestanley:
Looks better now but I don't think you needed to remove the trim test, I was just pointing out that you had something committed, that you didn't have in the description.
@CodeKrakken

lukestanley:
Ah I see! 'should parse Unix millisecond timestamp and return date string' and 'should parse Unix millisecond timestamp and return date string with format argument' both should not use numbers, but they wrongly do, it makes sense in other functions but not those.
@CodeKrakken

lukestanley (approved):
:+1: :sparkles:

Issue #299, opened 2022-06-23
Use greyed out buttons instead of hidden buttons


dahacouk:
Would it be possible to use greyed out buttons instead of hidden buttons for _Export_ and _Delete_ at https://app.kendra.io/schemas/list when no Schema is selected?

<img width="470" alt="image" src="https://user-images.githubusercontent.com/201752/175369637-f646c615-cdc5-4268-89f6-b32a9ad4cc93.png">

Happy to discuss if too much trouble. Cheers @dahacouk

the-greenman:
https://kendraio-lx3vvuhks-kendraio.vercel.app/schemas/list

dahacouk:
Hi Luke and Gio. Could we get this pushed into main please? Any blockers to that? @lukestanley @gsambrotta

gsambrotta:
Hi @dahacouk I did not see your comment, sorry. I will work on this one.

gsambrotta:
Probably related: https://github.com/kendraio/kendraio-app/pull/326/files

lukestanley:
> Hi Luke and Gio. Could we get this pushed into main please? Any blockers to that? @lukestanley @gsambrotta
This is an issue so it on it's own has nothing to push, but Gio has correctly identified the related PR for it, well done Gio. I've also linked the PR to this issue by mentioning it. Sounds like Gio will take up reviewing it. @dahacouk

PR #434, opened 2024-01-24
Actions button disabled instead of hidden

gsambrotta:
Actions blocks that are not selectable used to be hidden. 
Not they are disabled (grey out) and enabled once they are selectable.

UI test can be done on /schemas/list page

Related Issue #299


lukestanley:
Can this have a test with it please?

cypress/e2e/actions.ts
```diff
@@ -0,0 +1,78 @@
...
+    //     });
+
+    //     // Prevent external network requests for Workflow cloud
+    //     cy.intercept('GET', 'https://app.kendra.io/api/workflowCloud/listWorkflows', {
+    //         fixture: 'workflow-cloud.json'
+    //     });
+
+    //     // Prevent external network requests for fonts with empty CSS rule 
+    //     cy.intercept('https://fonts.googleapis.com/\*\*', "\*{ }");
+    // });
```
lukestanley, line 22: Did you mean to leave this commented? @gsambrotta

docs/workflow/blocks/actions.rst
```diff
@@ -38,27 +46,109 @@ This example shows two buttons. The first one dispatches an asynchronous command
...
+            "blocks": [
+                {
+                    "type": "launch",
+                    "valueGetters": {
+                        "adapter": "data.adapterName",
+                        "workflowId": "data.id"
+                    }
+                }
+            ],
+            "enabled": true // automatically added by the actions block. Not by a human.
```
lukestanley, line 93: If comments are in some JSON, it is likely to cause an error when pasted. The comment could go in the explanation text instead. @gsambrotta

docs/workflow/blocks/actions.rst
```diff
@@ -38,27 +46,109 @@ This example shows two buttons. The first one dispatches an asynchronous command
...
+        }
+    ]
+}
+
+
+enabledGetter
+++++++++++++++++++++++
+`enabledGetter` decided if the button is enabled or disabled. This is mostly referring to the UI of the button.
```
lukestanley, line 101:
```suggestion
`enabledGetter` determines if the button is enabled or disabled. This is mostly referring to the UI of the button. 
```

docs/workflow/blocks/actions.rst
```diff
@@ -38,27 +46,109 @@ This example shows two buttons. The first one dispatches an asynchronous command
...
+            "enabled": true // automatically added by the actions block. Not by a human.
+        }
+    ]
+}
+
+
+enabledGetter
+++++++++++++++++++++++
+`enabledGetter` decided if the button is enabled or disabled. This is mostly referring to the UI of the button. 
+In case the button is disabled, it will be displayed as grey out, and it will not be clickable.
```
lukestanley, line 102:
```suggestion
If the button is disabled, it will be displayed as grey out, and it will not be clickable.
```

docs/workflow/blocks/actions.rst
```diff
@@ -38,27 +46,109 @@ This example shows two buttons. The first one dispatches an asynchronous command
...
+`enabledGetter` expect a value that will ultimately return in a `true` or `false`.
```
lukestanley, line 103:
```suggestion
`enabledGetter` requires a JMESPath mapping value that should return `true` or `false`.
```

docs/workflow/blocks/actions.rst
```diff
@@ -38,27 +46,109 @@ This example shows two buttons. The first one dispatches an asynchronous command
...
+The value can be added manually or taken from the state or the context.
+In this example we want to display two different button, one enabled and another disabled.
```
lukestanley, line 105:
```suggestion
In this example we want to display two different buttons, one enabled and another disabled.
```

docs/workflow/blocks/actions.rst
```diff
@@ -38,27 +46,109 @@ This example shows two buttons. The first one dispatches an asynchronous command
...
+          "color": "default",
+          "blocks": [],
+          "enabledGetter": "state.global.disabled",
+          "enabled": false
+        }
+      ]
+    }
+
+
+After the block as loaded, the `enabled` property will be autogenerated:
```
lukestanley, line 130:
```suggestion
After the block as loaded, the `enabled` property will be set on the button:
```

docs/workflow/blocks/actions.rst
```diff
@@ -38,27 +46,109 @@ This example shows two buttons. The first one dispatches an asynchronous command
...
+          "label": "Disabled",
+          "color": "default",
+          "blocks": [],
+          "enabledGetter": "state.global.disabled",
+          "enabled": false
+        }
+      ]
+    }
+
+Buttons are not building to be used for external links. In that case, is better to use :doc:`Template Block <template>`
```
lukestanley, line 154:
```suggestion
Buttons do not work for external links. Consider the :doc:`Template Block <template>` for external links.
```

lukestanley (changes requested):
:sparkles:  This is really good overall. Just a few things need tweaking.

docs/workflow/blocks/actions.rst
gsambrotta: Ahh, you are right! Hmm, I did add the explanation in the text, but I liked to have it highlighted in here too. 
Well, I guess we cannot.

docs/workflow/blocks/actions.rst
```diff
@@ -26,8 +28,14 @@ Supported properties
...
 
   - **label** (string) - the text for the button label
+  - **id** (string) - the unique identifier for the button
   - **color** (string) - passed as the “color” attribute to the material button. Use one of the supported
     Material color values, such as “primary”, “warn”, “accent”. Leave as “default” to use the default button styling for a plain button.
+  - **enabledGetter** (string) - A value from the state or context. Based on this value, it generates an "enabled" property on the button.
+  The button will be shown as disabled (grey out) or enabled accordingly. The default value of this property is `true`. After 
+  the block is loaded, it will automatically add a property to the block json: `"enabled": true`.
+  If `enabledGetter` is set to `false` the automatic property will be then shown as following: `"enabled": false`.
+  - **valueGetters** (string) - A value that will be pass as context to whichever action is being executed.
```
lukestanley, line 38: I think this needs to point out that it's a JMESPath mapping, that is evaluated like the Mapping block is, to determine if the button should be enabled or not. @gsambrotta

CodeKrakken (changes requested):
Just a few suggestions on the syntax. Tests are all passing.

docs/workflow/blocks/actions.rst
gsambrotta: @lukestanley Please suggest a text to add. I don't know much about the `valueGetters`. The text I wrote describe my knowledge of it.

docs/workflow/blocks/actions.rst
```diff
@@ -14,6 +14,8 @@ Default config
         {
           "label": "Action",
           "color": "default",
+          "enabledGetter": true,
+          "enabled:" true,
```
lukestanley, line 18:
```suggestion
          "enabled": true,
```
:sweat_smile:

lukestanley: When this is fixed it'll be good to go I think. @gsambrotta

lukestanley (approved):
I'm approving this in advance. The only problem I found is the missing "enabled" property close quote in actions.rst! Great work! @gsambrotta

PR #442, opened 2024-01-30
Add GitHub Issue template

lukestanley:
Adds a template to help ensure we capture key details so we can reproduce them, know how important problems are, and so on.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,32 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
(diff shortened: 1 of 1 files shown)


gsambrotta (approved):
:clap: Looks great to me!

lukestanley:
> For example, [this random issue I found by Googling](https://github.com/Josee9988/project-template/issues/new?assignees=Josee9988&labels=Type%3A+Bug&projects=&template=1-bug-report.md&title=%5BBug%5D+), has helper text clearly marked. Am I missing something here?

Cool, I don't think you're missing something, I didn't realise that "comment block" syntax was supported! I've now updated the suggested issue template that now uses it. I used the template for a real issue, here: https://github.com/kendraio/kendraio-app/issues/444

dahacouk:
Cool @lukestanley. I know I'm nit-picking but how about removing the colons from the headings and being consistent? It helps nothing but promote the practice of being consistent – which we should all aspire to. Cheers!

gsambrotta (approved):

Issue #445, opened 2024-02-06
Fix Grid on Angular 16


gsambrotta:
While upgrading to Angular 16 the grid components don't look right.
It seems there is some problem with [ag-grid-angular](https://www.npmjs.com/package/ag-grid-angular)
The UI doesn't look like it is applied at all 
The launch buttons don't show up

I think no test is needed for this task.

**Angular 15**
User settings Grid
![image](https://github.com/kendraio/kendraio-app/assets/1055531/007543e7-7cf9-4c4c-acaf-12a8ed29daa7)

Flow Cloud Grid
![image](https://github.com/kendraio/kendraio-app/assets/1055531/1a7d0f21-aa23-45c7-b0c8-0f4352afa436)

**Angular 16**
User settings Grid
![image](https://github.com/kendraio/kendraio-app/assets/1055531/4a69b266-1e9c-4b48-a622-1aef324dde24)

Flow Cloud Grid
![image](https://github.com/kendraio/kendraio-app/assets/1055531/1d1b19ac-d358-45fe-a55b-b00bc3b0adda)

gsambrotta:
PR open #447

PR #447, opened 2024-02-08
Fix Ag Grid UI and custom component render

gsambrotta:
Connected Issue: #445 

What changed:
- Swap position for base CSS file and theme CSS file. This was preventing the proper theme style to being applied.
- Update `frameworkComponents` to `components` as per documentation. This was introduced in Ag Grid 27 but became a breaking change just in v29. 
- Update Ag Grid to last version 31, as now it is fully compatible with Angular 16.

How does it look like now:
image


CodeKrakken (approved):
This all looks fine, no comments. Tests passing.

lukestanley (approved):
Brilliant sleuthing Gio! @gsambrotta :female_detective: :sparkles: :+1:

PR #451, opened 2024-02-20
Convert "Workflow" text to "Flow"

gsambrotta:
Wherever we have a text, label or documentation that use the refer to the Kendraio Flow, we want to call them `Flow`.
In this PR I substituted all the text that was calling the Flow in another way like:

- Workflow
- flow
- flow block


CodeKrakken (changes requested):
Text looks fine - but we have a failing test - 
```
/kendraio-app/cypress/e2e/context_and_state.ts 
line 232 - 'should allow the disabling of the datagrid'```

gsambrotta:
Ah thanks @CodeKrakken, locally all the tests were passing. I will check it again

lukestanley (approved):
:+1:

lukestanley:
@gsambrotta 
For some reason this was merged into angular-16 instead of develop.
I feel bad I didn't catch it earlier.
But can you make a Pull Request for merging this into the right place?
I consider it blocking the Angular work.

Issue #440, opened 2024-01-29
Angular 16 Upgrade Issues


lukestanley:
- [x] #441
  - [x] Find if AG Grid broken in this branch #445 
  - [x] Material UI: different spacing and font style #446 
  - [x] Card component has broken UI  #446
  - [x] "Add Task" modal has different UI #446 
- [x] /connet page last link text font style and size #446
- [x] Review tests, are force clicks needed with new Cypress version? https://github.com/kendraio/kendraio-app/pull/457
- [x] Undo "Workflow" text to "Flow" - regression #448 
- [/] In Flow Cloud, flows don't show up at first render - This part it will be taken care of by the improvement on the reactivity of the whole app. To see PRs that deal with this, look for issues and PR with `state` label

lukestanley:
This was made as part of a test, see comment on #441.

gsambrotta:
This is probably the cause of most of our UI related issues: https://material.angular.io/guide/mdc-migration

gsambrotta:
All the issue listed in this Issue have been merged, so I am closing this issue

PR #457, opened 2024-03-06
Cypress: Remove force:true from click 

gsambrotta:
This PR take care of the task: `Review tests, are force clicks needed with new Cypress version?` from[ Angular 16 Issue](https://github.com/kendraio/kendraio-app/issues/440)

`force:click` is [a functionality of Cypress ](https://docs.cypress.io/api/commands/click#Options) and is present in all the versions. 
It is used when an element needs to be clicked, but it is hidden by another element. (For example, a menu that shows up when user over on another element)
As per [documentation](https://docs.cypress.io/guides/references/error-messages#Ignore-built-in-error-checking) and best practice, we should not overuse it: 
> Be careful with this option. It's possible to force your tests to pass when the element is actually not interactable in your application.

To avoid using it, there is no one-solution-fit-all, every situation must be evaluated differently. 
In our case, Cypress covers some material UI element with an invisible element like focus. Therefore, is enough to use a specific HTML attribute to target the element to be clicked. 
A `data-cy` attribute was added to the HTML element, so Cypress can easily spot the element to click, even if is hidden by other elements.

### DIFF
--- a/cypress/e2e/gosub.ts
+++ b/cypress/e2e/gosub.ts
@@ -24,7 +24,7 @@ describe('Subroutine workflow block', () => {
                 "workflowId": "madeUpFlowIDA"
             }
         ]);
-        cy.get('mat-toolbar > button mat-icon').contains('settings').click({force: true});
+        cy.get('[data-cy="toolbar-setting-button"]').click();
         cy.get("#mat-expansion-panel-header-0").click();
         // cy.pause();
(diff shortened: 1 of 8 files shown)


CodeKrakken (approved):
No syntax issues and all tests passing. Approving.

lukestanley (approved):
:+1:

PR #484, opened 2024-07-22
Map country areas

lukestanley:


src/app/blocks/map-block/map-block.component.ts
```diff
@@ -28,23 +29,92 @@ contributors, <a href="https://creativecommons.org/licenses/by-sa/2.0/">CC-BY-SA
...
+        } else {
+          markerIcon = icon({
+            iconSize: [25, 41],
+            iconAnchor: [13, 41],
+            iconUrl: '/assets/marker-icon.png',
+            shadowUrl: '/assets/marker-shadow.png'
+          });
+        }
+        
+        if (lat && long) {
```
gsambrotta, line 102: Is it reasonable to create an object without LAT and LONG? Aren't they indispensable information to work with a map?

gsambrotta (changes requested):
Nice work and very well documented. Thank you for this, Luke. 
Some minor comments here

CodeKrakken (approved):
Very nice work!

src/app/blocks/map-block/map-block.component.ts
lukestanley: A country is an area, and so is a LAT and LONG, it is just smaller, as a point.

gsambrotta: Honestly @lukestanley I do not understand this comment. Can you help me understand better what do you mean?
My question is: Can map object exist and be valid without a LAT and LONG? If yes, how come?

Issue #483, opened 2024-07-22
Map multiple country areas


lukestanley:
For the Stepping Stones project Pilot, we'd like the map to be able to accept country areas by specifying 3 letter ISO country codes in the data used by the Map block with the ability to customise the pin.

PR #486, opened 2024-08-13
Map country areas and makes map points optional if country is used

lukestanley:
Makes Map block able to accept a "country" property in the array of map points it processes, that show as an area, the values must be 3 letter ISO country codes.
Shading for area can be adjusted.
Custom map pin points may be set to an emoji flag.
Documentation has been removed from this branch.
A unit tests file for the block was added but is removed from this branch.

image

This should meet requirements for: https://github.com/kendraio/kendraio-app/issues/483

To help deliver the code changes faster for Donald to use from main, my existing pull request which code, tests and documentation are to be broken up, with these code changes to be pursued first. The rest need doing for this issue: https://github.com/kendraio/kendraio-app/issues/488

### DIFF
--- a/src/app/blocks/map-block/map-block.component.ts
+++ b/src/app/blocks/map-block/map-block.component.ts
@@ -1,7 +1,8 @@
 import {Component} from '@angular/core';
 import {BaseBlockComponent} from '../base-block/base-block.component';
-import {icon, latLng, marker, tileLayer} from 'leaflet';
+import {icon, latLng, marker, tileLayer, divIcon, geoJSON} from 'leaflet';
 import {get, isArray, set} from 'lodash-es';
+import * as DOMPurify from 'dompurify';
 
 @Component({
   selector: 'app-map-block',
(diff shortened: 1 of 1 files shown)


src/app/blocks/map-block/map-block.component.ts
```diff
@@ -28,23 +29,92 @@ contributors, <a href="https://creativecommons.org/licenses/by-sa/2.0/">CC-BY-SA
...
+  countryStyle = {
+    color: 'gray',
+    weight: 2,
+    opacity: 0.3,
+    fillColor: 'gray',
+    fillOpacity: 0.5
+  };
+
+  // In-memory cache for country GeoJSON data
+  countryCache: { [key: string]: any } = {};
```
CodeKrakken, line 46: Do we need to use an `any` here?

CodeKrakken (approved):
I've queried one thing but I'm happy to approve this either way - just curious really.

src/app/blocks/map-block/map-block.component.ts
lukestanley: The object is being mutated, so I instinctively typed it as 'any' to avoid any issues. A different format could be possible.

PR #525, opened 2025-02-27
AWS SigV4 Signed Authentication for HTTP block

lukestanley:


CodeKrakken (approved):
Code looks good, all tests passing. Good stuff!

src/app/blocks/http-block/http-block.component.ts
```diff
@@ -122,6 +125,57 @@ export class HttpBlockComponent implements OnInit, OnChanges {
...
+              if (k.toLowerCase() === 'host') return;
+              headers = headers.set(k, sigHeaders[k]);
+            });
+          } catch (err) {
+            this.hasError = true;
+            this.errorMessage = (err as Error).message;
+            this.isLoading = false;
+            return;
+          }
+          break;
```
gsambrotta, line 177: As far as I know, `break` here don't do anything. `break` in JavaScript exist just inside the switch block. 
Or am I missing something?

gsambrotta: ah, now I see the big switch above. All good then :)

gsambrotta (approved):
Looks good to me too.  Well done @lukestanley :tada:

PR #523, opened 2025-02-26
Json dereference block

gsambrotta:
- add block to Dereference JSON schemas 
- Fix Form component to be able to accept Jscon Schema as object, data or context

Issue #502, opened 2025-01-28
Add JSON Dereference block


gsambrotta:
**Task**
Create a new block which should take accept a JSON Schema document URL as input, and output a JSON Schema document as a string, or a JSON Schema object.

This would use dereferencing, a technique for merging referenced data from definitions that are stored in path addresses, into a merged schema.

**How does JSON Schema Dereferencing work?** 

JSON Schema document dereferencing returns a merged JSON Schema document by following the reference location (from the same document, or an external URL) and replacing the $ref with the resolved content of the definition directly in the schema.

1.  **Input**: Take JSON Schema URL string (content must point to a valid JSON schema)

2. **Output**: string or JSON Object

Example of using a JSON Schema Dereferencing library on a JSON Schema we tried to use:
 https://github.com/kendraio/kendraio-gists/blob/main/schema_merge_tool.ts

PR #527, opened 2025-03-12
JSON Dereference block, parser, documentation

gsambrotta:
PR related to Issue https://github.com/kendraio/kendraio-app/issues/502 

- add block to Dereference JSON schemas 
- Fix Form component to be able to accept Jscon Schema as object, data or context

You can test schema with string on this Flow: https://kendraio-app-git-json-dereference-block-kendraio.vercel.app/kendraio/schemaForm

You can test schema with blocks on this Flow:  https://kendraio-app-git-json-dereference-block-kendraio.vercel.app/kendraio/schemaFormContextBlocks

Here is a Flow with murmurations:
https://kendraio-app-git-json-dereference-block-kendraio.vercel.app/murmurations/loadSchema

Copy of https://github.com/kendraio/kendraio-app/pull/523 but based on Angular 16 rather then Angular 18


CodeKrakken (changes requested):
Just a couple of bits of syntax.

Caveat - I have not been able to run the tests.

lukestanley:
I think this would really benefit from some automated tests to help validate it. @gsambrotta

gsambrotta:
> Now you have tests passing, you can mock the resources they use. @gsambrotta

Why can't we keep the real life resources? So is even more realistic the test. It will let us know also if the Murmurations schema change or if the API fail @lukestanley

lukestanley:
> > Now you have tests passing, you can mock the resources they use. @gsambrotta
> 
> Why can't we keep the real life resources? So is even more realistic the test. It will let us know also if the Murmurations schema change or if the API fail @lukestanley
None of the existing Cypress tests are intended to be end-to-end service integration tests, they are intended only to tell us when our own app code is broken. Our tests take long enough to run, they should run as fast as reasonably possible. That's why we have mock request intercepts.

I fully agree that doing service integration tests would indeed be _a good thing_, but that should be a **separate** thing. Do you agree that it is better for our _sanity_, to have consistent tests? My opinion is that symmetry and clean patterns can really help, I am biased towards thinking that the less we have to fit in our heads or the less documentation we have to read about quirks, having unexpected gotchas is better? It's already quirky enough :)

CodeKrakken (approved):
Works for me! Good work @gsambrotta .

PR #534, opened 2025-05-14
Feat: jsonParse tool in mapping block

gsambrotta:
Adding JSON.parse utility to Mapping Block.

In this way, we can easily transform a JSON String to a workable Object.

How to use it in a Flow:
image

Documentation Updated:
image

### DIFF
--- a/docs/workflow/blocks/mapping.rst
+++ b/docs/workflow/blocks/mapping.rst
@@ -295,17 +295,18 @@ Available Extensions
 19. zip
 20. debug
 21. json
-22. markdown
-23. btoa
-24. base64encode
-25. pairwise
-26. numDiff
-27. percentChange
-28. groupByKeys
-29. all
-30. parseDate
-31. parseDuration
-32. parseUnixTimestamp
+22. jsonParse
+23. markdown
+24. btoa
+25. base64encode
+26. pairwise
+27. numDiff
+28. percentChange
+29. groupByKeys
+30. all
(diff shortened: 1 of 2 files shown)


lukestanley:
This might make sense from the mapping block, but did you notice the parse data block already exists?
https://kendraio-app.readthedocs.io/en/latest/workflow/blocks/parse_data.html

CodeKrakken (changes requested):
Looks good, couple of comments though.

src/app/blocks/mapping-block/mapping-util.spec.ts
```diff
@@ -86,4 +86,12 @@ describe('MappingUtil', () => {
...
     expect(mappingUtility(data, expr)).toBe(expected);
   });
+
+  it('should parse a valid JSON string into an object', () => {
+    const data = { jsonString: '{"name": "Alice", "age": 30}' };
+    const expr = "jsonParse(jsonString)";
+    const expected = { name: "Alice", age: 30 };
+
+    expect(mappingUtility(data, expr)).toBe(expected);
+  });
```
CodeKrakken, line 96: This test throws an error unfortunately - 

`Expected Object({ name: 'Alice', age: 30 }) to be Object({ name: 'Alice', age: 30 }). Tip: To check for deep equality, use .toEqual() instead of .toBe()`

Fortunately that's very clear! I checked, and the change resolves the error.

PR #537, opened 2025-06-17
Update dependencies

gsambrotta:
Upgrading dependencies.

1 Step: Upgrade all patch versions. Angular packages to 18.1.5


CodeKrakken (approved):
No problems here - all dependencies install correctly and all tests pass. Nice work!

PR #545, opened 2025-07-15
Upgrades Cypress container to 14.5.1

### LINKED ISSUE #544: Upgrade Cypress test container version
When GitHub runs our tests, currently in the `.github/actions/tests/Dockerfile` file, we use `cypress/included:8.4.1` as the Docker container image, which is based on the Debian Buster Linux distribution, which now breaks while the container image is being built and so it cannot access updated packages from Debian's server, as shown by this error in the test logs:

[`E: The repository 'http://deb.debian.org/debian buster Release' no longer has a Release file.`](https://github.com/kendraio/kendraio-app/actions/runs/16270232069/job/46013693180#step:3:226)

lukestanley:
**Fixes broken tests caused by old Cypress container**

Our GitHub Actions tests started failing because the Cypress Docker image was based on Debian Buster, which has reached end-of-life, and the package repositories are no longer readily accessible.
The container runs our unit tests and the end-to-end Cypress tests.
This upgrades the container from `cypress/included:8.4.1` to `cypress/included:14.5.1`, which uses a supported Debian version and already includes Node.js v22.17.

**Changes:**
- Upgrades Cypress container to 14.5.1
- Removed redundant Node.js 18 installation (now included)
- Added `xauth` package (required but no longer bundled)

Fixes #544

### DIFF
--- a/.github/actions/tests/Dockerfile
+++ b/.github/actions/tests/Dockerfile
@@ -1,15 +1,7 @@
-FROM cypress/included:8.4.1
-
-# Install Node 18
-RUN curl -fsSL https://deb.nodesource.com/setup_18.x | bash - && \
-    apt-get install -y nodejs && \
-    mv /usr/local/bin/node /usr/local/bin/node_old && \
-    ln -s /usr/bin/node /usr/local/bin/node && \
-    npm install -g npm@latest && \
-    node --version && npm --version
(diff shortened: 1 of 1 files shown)


CodeKrakken (approved):
That all looks good, all tests passing. Good stuff!

PR #548, opened 2025-07-25
Allow Mapping Block json() util to take arguments

gsambrotta:
Improve mapping `json()` utility function so that can take arguments.
This is done in order to render a prettified JSON output as user needs to copy and paste it.

DONE:
- allow json() to accept args
- write unit test for json() 
- update documentation about json(), qs() and jsonParse() so is clear the difference.

`json(data)` BEFORE:

image
image (1)

`json(data, null, 2)` AFTER - using args:

image (2)

### DIFF
--- a/docs/workflow/blocks/mapping.rst
+++ b/docs/workflow/blocks/mapping.rst
@@ -609,6 +609,16 @@ Use json-stringify-safe to safely stringify and accept arguments for replacer an
 Examples:
 
 .. code-block:: javascript
+
+   json({
+      "name": "Alice",
+      "tags": [
+         "dev",
+         "ui"
+      ]
+   })
+
+   // Output on one line: "{{"name": "Alice","tags": ["dev","ui"]}"
    
    json({
       "name": "Alice",
@@ -618,7 +628,7 @@ Examples:
       ]
    }, null, '2')
 
-   // Output: "{
+   // Output prettified: "{
 .. {
 ..   "name": "Alice",
 ..   "tags": [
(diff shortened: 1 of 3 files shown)


CodeKrakken (approved):
That's all good, code is clear and all tests passing. Nice!

lukestanley:
@gsambrotta Thanks for disambiguation from jsonParse, I don't recall agreeing, but I see that you merged it to develop. I wasn't quite sure what the motivation for it was. This feels very similar but a different direction.

Looking at your snippet and explanation,  are you sure that is correct?
Surely the 'as' in the import does a rename to avoid clashing with a _different_ stringify function from json-stringify-safe, which is **not** related to query strings?

Is this blocking any Flow work?

I don't see an attached GitHub issue.

Is there an agreed problem this Pull aimed at solving?

I think we need to get better at having issues to scope the work we do.

lukestanley (changes requested):
It's not clear that this is a useful, well scoped change that solves a real problem.

gsambrotta:
@lukestanley I have updated the code and the PR description as agreed in the meeting. 
Let me know if you see any issues.

lukestanley:
I think a test for the existing functionality would be good, not just the new behaviour.
We noticed in the meeting that it isn't branching off from main yet but not many files are touched so it should be straight forward to reapply.
I'm not quite sure the document change makes sense yet.

lukestanley:
Your code changes the signature, so it would be good to have a test showing that using the old signature still works as expected too, so we know we have coverage, right?

The documentation changes don't reuse the same shared data object as other examples, they introduce new data and add a JavaScript comment to direct the user to it.
It might be better to reuse the existing common data object, if every example had introduced a new data input and new way of referring to it, it would not be very harmonious or easy to follow.

It might be worth replacing the whole thing with a snapshot test and documentation generating combination at some point that just has one shared data / context setup step and an array of example JMESPath queries for each method. Using different data makes that harder.

gsambrotta:
PR against main can be found here: https://github.com/kendraio/kendraio-app/pull/550

Issue #566, opened 2025-10-01
Show app modification date


lukestanley:
- Service workers have cached versions of Kendraio App's static files.
- [Old service workers are being used unexpectedly, intermittently.](https://github.com/kendraio/kendraio-app/pull/538#issuecomment-3355782338)
- Seeing the modification date of the app service worker will help us know what version we are using.

Requirement:
- Show the service worker modified date in the left sidebar footer 
- Show the modified date in the format of `YYYY-MM-DD HH:MM:SS UTC`.

Currently out of scope:
- Clearing service worker cache / state.

PR #569, opened 2025-10-10
Show app modified date

lukestanley:
Shows the modified date at the bottom of the left sidebar.
Implementing #566

### DIFF
--- a/src/app/services/service-worker-info.service.ts
+++ b/src/app/services/service-worker-info.service.ts
@@ -1,8 +1,8 @@
 import { Inject, Injectable, PLATFORM_ID } from '@angular/core';
 import { HttpClient, HttpHeaders } from '@angular/common/http';
 import { isPlatformBrowser } from '@angular/common';
-import { Observable, defer, of } from 'rxjs';
-import { catchError, map, shareReplay, switchMap } from 'rxjs/operators';
+import { Observable, of } from 'rxjs';
+import { catchError, map } from 'rxjs/operators';
 
 export interface ServiceWorkerStatus {
   readonly lastModified: Date | null;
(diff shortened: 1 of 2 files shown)


CodeKrakken (approved):
Lovely stuff @lukestanley , no comments.

src/app/app.component.scss
```diff
@@ -0,0 +1,58 @@
...
+
+:host ::ng-deep mat-sidenav-container.app-container {
+  display: flex;
+  flex: 1 1 auto;
+  height: 100%;
+  min-height: 0;
+  position: relative;
+}
+
+/* Ensure the sidenav itself stretches full height and behaves as a column */
```
gsambrotta: We don't usually write all these comments. 
in my opinion, some might be useful, but most of them explain rather basic CSS styling. 
Anyway, I think we should stay consistent with the rest of the code and don't introduce over explanatory comments. What do you think @lukestanley ?

src/app/services/service-worker-info.service.ts
```diff
@@ -0,0 +1,64 @@
...
+        })
+      : undefined;
+
+    const url = bypassCache ? '/ngsw.json?ngsw-bypass=true' : '/ngsw.json';
+
+    return this.http.get<{ timestamp?: number }>(url, { headers }).pipe(
+      map(manifest => {
+        const timestamp = manifest?.timestamp;
+        if (typeof timestamp !== 'number' || !Number.isFinite(timestamp)) {
+          return { lastModified: null };
```
gsambrotta: We probably want a more helpful output here, something like: throw new Error('Invalid timestamp in manifest'); or { lastModified: null, error: 'Invalid timestamp in manifest' }

gsambrotta (commented):
Well done @lukestanley 👏🏻 
I added some small comments.

I have one concern: I find the service-worker-info.service.ts complicate to read. 
I think most of the complication come from the heavy usage of RxJS and the concatenation of observables.
Is there any specific reason why you used Observable rather than async/await?

gsambrotta:
Sorry I forgot to review it yesterday. Anyway I had a look now at the async/await code and it looks good. 👍🏻 from me too

PR #581, opened 2025-11-19
Adds Flow block configuration and data to debug errors

lukestanley:
Currently when there a Block in a Flow has an error, the error information shown in the console lacks important context about which Block causes the error (see issue #582). 
This Pull Request adds recent Block configuration and data inputs to the browser console logs to help solve this problem.

### DIFF
--- a/src/app/components/blocks-workflow/blocks-workflow.component.ts
+++ b/src/app/components/blocks-workflow/blocks-workflow.component.ts
@@ -1,12 +1,17 @@
-import {Component, EventEmitter, Input, OnInit, Output} from '@angular/core';
+import {Component, EventEmitter, Input, OnChanges, OnDestroy, OnInit, Output, SimpleChanges} from '@angular/core';
 import {clone} from 'lodash-es';
+import {
+  registerWorkflowDebugProvider,
+  unregisterWorkflowDebugProvider,
+  WorkflowDebugSnapshot
+} from 'src/app/services/global-error-handler.service';
 
 @Component({
   selector: 'app-blocks-workflow',
   templateUrl: './blocks-workflow.component.html',
   styleUrls: ['./blocks-workflow.component.scss']
 })
-export class BlocksWorkflowComponent implements OnInit {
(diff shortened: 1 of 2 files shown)


src/app/components/blocks-workflow/blocks-workflow.component.ts
```diff
@@ -1,28 +1,60 @@
...
+  }
+
   updateModel(modelNumber, value) {
     // console.log({ modelNumber, value });
     // console.log(this.models);
     this.models = [...this.models.slice(0, modelNumber), value, ...this.models.slice(modelNumber + 1)];
+    const emittingIndex = modelNumber - 1;
+    this.lastActiveBlockIndex = emittingIndex >= 0 ? emittingIndex : null;
+    this.lastTargetBlockIndex = modelNumber;
+  this.trackActiveBlock(this.lastActiveBlockIndex);
```
CodeKrakken: Petty indentation comment

CodeKrakken:
> > @lukestanley Nice neat code aside from one indenting oversight. However there are a couple of failing tests -
> > S3 Bucket Binary Integrity Test
> > ```
> > * should upload the asset, retrieve it, and verify its integrity
> > 
> > * should find image in bucket
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I believe these two tests have failed unexpectedly before. I tested on my laptop and desktop, twice each with a totally fresh install of the app.
> 
> @CodeKrakken But those same tests are passing in CI here with no error, so I suspect you are not running the s3rver mock before running the Cypress tests that depend on them? I also ran it fine locally. I have fixed the indent error however, good catch!

@lukestanley Ah yes that must be it, I didn't run the s3rver mock. How does one do that?

I think if possible that the tests should run s3rver if they need it - otherwise a note on the PR would be handy!

CodeKrakken:
@lukestanley Just retested with s3rver running - all tests passing now - good stuff.

